PR #221, opened 2015-02-11
Feature/jel 215 admin settings overrideable

### LINKED ISSUE #215: Make admin panel settings to be set via DB or ENV
Need to make admin settings configurable via DB or environment variables.

OBCENEIKON:
Fixes #215


stackus:
This take a completely different approach.

- Add disabled to setting_fields
- Create an application initializer
  - Load up ENV values that match your specified override pattern
  - Process every setting, default to `disabled: false` for all settings (this will reset any previously disabled settings), if there is a override match, set `value: override` and `disabled: true`
  - Prior to saving set `override_disabled: true` for the record
- Add `attr_accessor :override_disabled` to SettingField
- Add a before_save hook where you have `false if disabled? && !override_disabled`

OBCENEIKON:
Hey Michael,

So after a time of figuring out initializers and working on your suggestions, I think I have what you’re wanting, though it’s not working…

I created an initializer file:
```
# config/initializers/env_settings.rb

class EnvSettingsRailtie < Rails::Railtie
  config.to_prepare do
    ActiveRecord::Base.connection
    SettingFields = ActiveRecord::Base::SettingField.all

    return if SettingFields.nil? || ENV.nil?

    SettingFields.each do |setting_field|
      setting_field.disabled = false

      unless setting_field.env_var_name.nil? || ENV[setting_field.env_var_name].nil?
        setting_field.override_disabled = true
        setting_field.value = 'override'
        setting_field.disabled = true
        setting_field.save
      end
    end
  end
end
```
Then in the setting_field model:
```
# app/models/setting_field.rb

class SettingField < ActiveRecord::Base
  belongs_to :setting

  store_accessor :options

  enum field_type: [:check_box, :select_option, :text, :date, :password]

  attr_accessor :override_disabled
  before_save :override_disabled_check

  private

  def override_disabled_check
    false if disabled? && !override_disabled
  end
end
```
The environment variable's value is never being passed to the UX and we’re overwriting the value stored with “override.” I understand why the value isn’t being passed to the UX and why it’s being overwritten, but I am not understanding why you wanted this and how it’s supposed to work.

We need some sort of way to intercept a call to the DB and replace the value with the environment variables value or prevent it from being saved. Using the before_save hook to return false if the environment variable is set did the trick and using an after_find hook also worked to replace the value.

stackus:
The line `setting_field.value = 'override'` should be setting the value to
the value of the ENV variable not a string that reads 'override'.

Also save yourself some boolean logic headache and only fetch the records
from the database that have something in the env_var_name column e.g.
`where.not(:env_var_name => nil)`. Now you only have to check if
`ENV[setting_field.env_var_name].empty?` on that 'if' the ||.

The return if line will never return. If the app is running in a place
where ENV is unavailable you're in trouble, and `.empty?` should be used
for the results list.

Don't use capitalized variables in ruby. Capitalized names are constants.

OBCENEIKON:
I originally had it set to the ENV variable, but this is going to save that ENV var to the DB, which I don't think Chris was wanting that to be done.

I was also originally only fetching any setting_fields that had an env_var_name, but then decided to fetch all after reading your comment "Process every setting_field, default to ```disabled: false``` for all settings (this will reset any previously disabled settings), if there is a override match, set ```value: override``` and ```disabled: true```"

Let me verify with Chris that we are not wanting to have the values saved to the DB.

OBCENEIKON:
Ok, that is correct, we don't want any values in the environment variables to be saved to the database.

stackus:
Right about the getting all to clearing disabled. The code shown wouldn't
clear the disabled flag for any setting because the save was happening
within the check for env_var_name.

If he doesn't want values saved into the database I think this might need
some more thought.

stackus:
Then these are not settings. They aren't configurable and are not variable
any longer. These are application configuration values. Once again, the ENV
isn't variable while the application is running so changing the ENV values
mid-run won't fix, connect, or change anything.

My opinion is : If these are settings that should never be sent to a
client, should never be changed by a client then save everyone the effort
and just pull them from the environment and be done with it. There is
nothing gained by any client using the API that there's a value they cannot
change and can't know about. Why even tease them?

PR #246, opened 2015-02-17
Added total to the order_history payload. Fixes #243.

### LINKED ISSUE #243: On /project/:id under "PROJECT ORDERS" all orders are $0.00
On /project/:id under "PROJECT ORDERS" all orders are $0.00.  Please investigate and fix so it is correct dollar amount.  The dollar amount is defined as the 1st month charge (setup + monthly + 750 hours of system)

OBCENEIKON:


app/models/project.rb
```diff
@@ -54,6 +54,7 @@ def order_history
     history = Order.where(id: OrderItem.where(project_id: id).select(:order_id)).map do |order|
       order_json = order.as_json
       order_json[:item_count] = order.item_count_for_project_id(id)
+      order_json[:total] = monthly_spend
```
stackus: `monthly_spend` is the total for the project, not an order. And it should already be returned as an attribute off of the project object.

Issue #254, opened 2015-02-18
Sample rake file was not encrypting authentication_token


OBCENEIKON:

stackus:
Also, the sample file should use `secret: '{token here}'` and not hardcode the pre-encrypted token.

mafernando:
@stackus - where should the secret come from? randomly generated or set in .env beforehand?

stackus:
@mafernando The sample uses `password: 'jellyfish'`, so `secret: 'jellyfish-token'` works. I meant encoded token, not pre-encoded token. (switched to tea today).

That said, I don't have an example of how the encrypted password or authentication token being in the sample could lead to a system hack. I do feel that putting encrypted strings into a repo is a bad thing™.

PR #255, opened 2015-02-18
Updated the sample.rake file with the authentication_token.

OBCENEIKON:
Setting it to jellyfish-token for now, just need to get it in there for a demo.


OBCENEIKON:
Fixes #254

Issue #109, opened 2015-01-21
Add ability to use SSO


chriskacerguis:
Add ability to use SSO to login to Jellyfish.

PR #265, opened 2015-02-21
Feature/109 sso authentication

stackus:
#109 
This PR has a matching PR in UX merge them both after comments and approval.

Merge before https://github.com/projectjellyfish/ux/pull/209.


app/controllers/settings_controller.rb
```diff
@@ -1,7 +1,7 @@
 class SettingsController < ApplicationController
   before_action :load_settings, only: [:index]
   before_action :load_setting, only: [:update, :destroy, :edit]
-  before_action :load_setting_by_name, only: [:show]
+  before_action :load_setting_by_hid, only: [:show]
```
chriskacerguis: @stackus Could you just use the "env_var_name" column?

stackus: I decided not to go with the `env_var_name` for several reasons. It wasn't clear that every setting would have a value in this column, and that the purpose of the column is to connect API with the underlying system so there's a chance they might change if a situation arose where a given name was taken or should be changed.

Also exposing system setup, environment var names and values, could be seen as a security risk so relying on them for another purpose like client lookups might be risky.

chriskacerguis: @stackus trying to understand things here.  

No there is not a "env_var_name" on the Settings table, however there IS on the settings_fields table.  

Based on what I see prior to your commit, there is an "env_var_name" on every row.

Regarding "exposing system setup, environment var names and values", could you explain what you mean by that?  Since the project is open source, the ENV vars are public (and they would need to be so people could use them).  

Not trying to argue, just trying to understand.  :)

stackus: What I meant was setting and using a value in the servers environment is typically done for security reasons. Essentially someone would have to had gained access to the box to get to the value. Things such as external API credentials, login user names and passwords, database server information and so on. Exposing this information the way API does, by taking values that have been set in the environment and putting them into responses sent to any authenticated client asking for them removes any advantage of security by using environment variables.

It was not my understanding that every setting would have a value in `env_var_name`, just that they have have so far. The purpose of a human id is be a 'unique unchanging id for lookup'. While they have similar purposes the `env_var_name` column could be changed if it had to, and I felt forcing a 'unique unchanging' rule on it shouldn't be done.

PR #283, opened 2015-02-26
fix demo seed file

mafernando:
remove ip_address from rake sample demo to align with order item ip attribute change


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/feature/demo-seed-add-order-item-ip-change">analyzed this pull request</a>.

stackus:
You could have also renamed the column to match it's new name from the migration it was renamed in.

mafernando:
@stackus - two columns were added to replace ip_address: public_ip and private_ip. Thought it would be simpler to just leave exclude both columns as ip_addresss was set to nil in the seed file anyway.

stackus:
I did see that. Either approach is fine, I prefer to not make edits that remove data, even nils.

Issue #275, opened 2015-02-24
'protect_from_forgery' should be called in ApplicationController


chriskacerguis:
A  vulnerability was found by Code Climate.

https://codeclimate.com/repos/54d4fa64695680762a0024c5/security/a88969674cc5a97cd65b06a0494358fb

chriskacerguis:
Can you please investigate this?  I know that we been messing with that on and off.

stackus:
API+UX is an odd duck. API+UX create a single site like experience but it's really an API with an HTML5 client.

chriskacerguis:
I think your PR from a few minutes ago resolved this issue

PR #286, opened 2015-02-26
Include protect_from_forgery

stackus:
- Remove the csrf gem. It does nothing for api. UX and other clients will not be running on the same domain (or port) as API.
- Skip csrf check if the request is a json request (APIs primary communication method)

#275

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -92,9 +92,6 @@ gem 'virtus'
 # CORS
 gem 'rack-cors'
 
-# CSRF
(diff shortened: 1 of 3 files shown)


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/feature/275-protect-from-forgery">analyzed this pull request</a>.

PR #293, opened 2015-02-27
Feature/jel 209 fog integration

jntullo:
- AWS Fog integration
- Includes mocking
- Updates to sample data


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/feature/jel_209_fog_integration">analyzed this pull request</a>.

chriskacerguis:
@jntullo Can you take a look at CodeClimate, and see if you can clean up some of the areas.

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,157 @@
+class AwsFog
+  def initialize(order_item_id)
+    @order_item_id = order_item_id
+    ENV['MOCK_MODE'] == true ? Fog.mock! : Fog.unmock!
```
stackus: Is this something that is true only in development? Is there a use for this in a production install?

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,157 @@
...
+    details
+  end
+
+  # TODO: Need to come up with a better way to manage CamelCase vs snake_case for DB instance creation
+  def order_item_details_camelize
+    details = {}
+    answers = order_item.product.answers
+    order_item.product.product_type.questions.each do |question|
+      answer = answers.select { |row| row.product_type_question_id == question.id }.first
+      if question.manageiq_key == 'db_instance_class'
```
stackus: You can and should change these keys to what manageiq needs/expects, I originally used snake case because that seemed reasonable.

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,157 @@
...
+    )
+    details = order_item_details
+    # TODO: Must get an image_id from product types
+    details['image_id'] = 'ami-acca47c4'
+    begin
+      create_infrastructure(details)
+    rescue Excon::Errors::BadRequest
+      order_item.provision_status = :critical
+      order_item.status_msg = 'Bad request. Check authorization credentials.'
+      order_item.save
```
stackus: You can add an `ensure` block and put `order_item.save` there instead of duplicating it.

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,157 @@
...
+    rescue ArgumentError => e
+      order_item.provision_status = :critical
+      order_item.status_msg = e.message
+      order_item.save
+    rescue StandardError => e
```
stackus: You can combine these two rescues like this `ArgumentError, StandardError -> e`

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,157 @@
...
+      order_item.status_msg = e.message
+      order_item.save
+    end
+  end
+
+  def create_database(details, db_instance_id)
+    db = @aws_connection.create_db_instance(db_instance_id, details)
+    order_item.provision_status = :ok
+    order_item.username = 'admin'
+    order_item.password = BCrypt::Password.create(@sec_pw)
```
stackus: @sec_pw does not exist.

I would pass in the sec_pw to this function instead of making it a member variable though.

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,157 @@
...
+    end
+    details
+  end
+
+  def aws_settings
+    @aws_settings ||= Setting.find_by(hid: 'aws').settings_hash
+  end
+
+  def provision_infrastructure
+    @aws_connection = Fog::Compute.new(
```
stackus: This can be put into a method with a similar signature as `aws_settings`.

```
def aws_connection
  @aws_connection ||= Fog::Compute.new....
end
```

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,157 @@
...
+    order_item.provision_status = 'ok'
+    order_item.save
+  end
+
+  def provision_databases
+    @aws_connection = Fog::AWS::RDS.new(
+      aws_access_key_id: aws_settings[:access_key],
+      aws_secret_access_key: aws_settings[:secret_key]
+    )
+    sec_pw = SecureRandom.hex[0..9]
```
stackus: `SecureRandom.hex 5`

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,157 @@
+class AwsFog
+  def initialize(order_item_id)
+    @order_item_id = order_item_id
+    ENV['MOCK_MODE'] == true ? Fog.mock! : Fog.unmock!
+  end
+
+  def provision
```
stackus: You could move the `begin...rescue...ensure...end` blocks into here and `case` on the types currently supported. Most of the method appear to setup some values and then call a `create_*` method. Some of that setup can either be done before the `case` or within the `create_*` methods. Doing this should make code climate happy.

app/workers/aws_fog.rb
jntullo: So an issue I ran into that I need to look into further is that the AWS sdk (used in MIQ) uses snake case, but fog for this instance only (RDS) is using camelcase.

app/workers/aws_fog.rb
chriskacerguis: @stackus This would also also be enabled on production DEMO installs.

app/workers/aws_fog.rb
stackus: OK, I just wanted to be sure no one felt compelled to use the names I entered initially.

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,113 @@
...
+  def initialize(order_item_id)
+    @order_item_id = order_item_id
+    ENV['MOCK_MODE'] == 'true' ? Fog.mock! : Fog.unmock!
+  end
+
+  def provision
+    product_type = order_item.product.product_type.name.capitalize.downcase
+    method_call = method("provision_#{product_type}")
+    begin
+      method_call.call
```
stackus:
```
send "provision_#{product_type}".to_sym
```

Then add a rescue for `NoMethodError` to catch the cases where the product_type isn't supported. You'll also be able to drop the `method_call = ...` line as well.

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,113 @@
+class AwsFog
+  def initialize(order_item_id)
+    @order_item_id = order_item_id
+    ENV['MOCK_MODE'] == 'true' ? Fog.mock! : Fog.unmock!
```
stackus: It occurs to me that this might not leave things in the right state. When a worker is created it gets marshaled into a yaml string after initialize is called. The check to switch Fog one way or the other is going to be left in the state the last worker initialized set it as.

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,113 @@
...
+    method_call = method("provision_#{product_type}")
+    begin
+      method_call.call
+    rescue Excon::Errors::BadRequest
+      order_item.provision_status = :critical
+      order_item.status_msg = 'Bad request. Check authorization credentials.'
+    rescue Fog::Compute::AWS::Error => e
+      order_item.provision_status = :critical
+      order_item.status_msg = e.message
+    rescue ArgumentError, StandardError => e
```
stackus: Looks like the `Fog::Compute::AWS::Error` can join this party.

app/workers/aws_fog.rb
```diff
@@ -0,0 +1,113 @@
...
+    sec_pw = SecureRandom.hex 5
+    details = order_item_details_camelize
+    # TODO: Figure out solution for camelcase / snake case issues
+    details['MasterUserPassword'] = sec_pw
+    details['MasterUsername'] = 'admin'
+    db_instance_id = "id-#{@order_item.uuid[0..9]}"
+    db = aws_connection.create_db_instance(db_instance_id, details)
+    order_item.provision_status = :ok
+    order_item.username = 'admin'
+    order_item.password = BCrypt::Password.create(@sec_pw)
```
stackus: Still have @sec_pw, and you probably want `sec_pw`.

app/workers/aws_fog.rb
jntullo: @stackus this is true. How do you suggest we go about this then? I just assumed we would restart the worker, but I guess that is an incorrect assumption.

stackus: You could put the ENV... line as the first line in `provision`. The `@order_item_id` will get serialized.

stackus:
:+1:

PR #297, opened 2015-03-02
Create Bundles of Products

calebthompson:
When a Bundle is added to an Order, the Order creates OrderItems with the
appropriate product/project pair.

* Add admin API for create/update of Bundles

### DIFF
--- a/app/models/bundle.rb
+++ b/app/models/bundle.rb
@@ -1,0 +1,21 @@
+# == Schema Information
+#
+# Table name: bundles
+#
+#  id          :integer          not null, primary key
+#  name        :string(255)
+#  description :text
+#  img         :string(255)
+#  active_at   :datetime
+#  deleted_at  :datetime
+#  created_at  :datetime
+#  updated_at  :datetime
+#
+
+class Bundle < ActiveRecord::Base
+  has_many :bundled_products
+  has_many :products, through: :bundled_products
+
+  time_for_a_boolean :deleted
+  time_for_a_boolean :active
+end
--- a/app/models/bundle.rb
+++ b/app/models/bundle.rb
@@ -13,9 +13,9 @@
 #
 
 class Bundle < ActiveRecord::Base
+  acts_as_paranoid
   has_many :bundled_products
   has_many :products, through: :bundled_products
 
-  time_for_a_boolean :deleted
   time_for_a_boolean :active
 end
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -28,6 +28,12 @@ class Order < ActiveRecord::Base
 
   accepts_nested_attributes_for :order_items
 
+  def bundle_id=(bundle_id)
+    order_items << Bundle.find(bundle_id).products.map do |product|
+      OrderItem.new(product: product, project: project)
+    end
+  end
+
   def item_count
     order_items.count
   end
@@ -42,5 +48,11 @@ class Order < ActiveRecord::Base
       total = total + order_item.setup_price + order_item.monthly_price + (order_item.hourly_price * 750)
     end
     total
+  end
+
+  private
+
+  def project
+    order_items.any? && order_items.first.project
   end
 end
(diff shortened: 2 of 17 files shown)


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/bundles">analyzed this pull request</a>.

calebthompson:
@sonejah21 It sounds like you might be the person to chat with about the front end for this. Is there anything obviously missing here?

app/models/bundle.rb
```diff
@@ -0,0 +1,21 @@
...
+#  deleted_at  :datetime
+#  created_at  :datetime
+#  updated_at  :datetime
+#
+
+class Bundle < ActiveRecord::Base
+  has_many :bundled_products
+  has_many :products, through: :bundled_products
+
+  time_for_a_boolean :deleted
```
stackus: Using the paranoia gem for this would give us the same helpers and also add a scope which excludes deleted rows.

calebthompson: Neat.

spec/models/setting_spec.rb
```diff
@@ -4,7 +4,7 @@
 #
 #  id   :integer          not null, primary key
 #  name :string(255)
-#  hid  :string(255)      not null
+#  hid  :string(255)
```
stackus: You'll want to revert this. An annotation gem issue is causing this file to get changed. The affected migration has been fixed.

app/models/bundle.rb
calebthompson: Looks like it won't work for active. Can you confirm that?

db/schema.rb
```diff
@@ -312,7 +332,7 @@
     t.integer  "setting_id"
     t.string   "env_var_name"
     t.boolean  "disabled",               default: false
-    t.string   "hid",                                    null: false
+    t.string   "hid"
```
stackus: Odd, the annotation issue shouldn't be affecting the schema. Perhaps merging in master will clear this up.

app/models/bundle.rb
stackus: Correct. You'll still use `time_for_a_boolean` for active.

db/schema.rb
calebthompson: I just did a migration. Let me do a fresh db build to make sure we're in sync with the migrations.

calebthompson: (And I'd rebased onto master)

sonejah21:
@calebthompson - looks good to me!

calebthompson:
Cool. @stackus care to take another glance? I'm adding API docs for BundlesController now.

app/models/order.rb
```diff
@@ -28,6 +28,12 @@ class Order < ActiveRecord::Base
 
   accepts_nested_attributes_for :order_items
 
+  def bundle_id=(bundle_id)
+    order_items << Bundle.find(bundle_id).products.map do |product|
+      OrderItem.new(product: product, project: project)
```
stackus: If `project` returns false, does `project: false` cause any issues? I honestly can't recall if I've ever tried sending in false to a relationship so I don't know.

calebthompson: I don't think project can ever be false, but it may be `nil`, which is somewhat problematic. I don't know that I have a good solution for that as Order params don't themselves include the project_id (as I think they probably ought to).

calebthompson: Problematic because it means we have an OrderItem without a project.

calebthompson: @stackus ?

stackus: `project` has `order_items.any? && order_items.first.project`. If there are no order_items `.any?` shortcuts the statement and will return false.

calebthompson: Truth.
```ruby
OrderItem.new(project_id: false).project_id
=> 0
```

calebthompson: Fixed in a962e9c

stackus: I ran a similar test with `OrderItem.new(project: false).project_id` and got `nil`. So when you pass in `false` to a relationship you'll get `nil` out the other end.

All the same, your final edit seems much safer.

:+1:

PR #303, opened 2015-03-03
Feature/jel 277 update order items table

OBCENEIKON:


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/feature/JEL-277-Update_order_items_table">analyzed this pull request</a>.

db/migrate/20150227033611_create_provision_derivations.rb
```diff
@@ -0,0 +1,12 @@
+class CreateProvisionDerivations < ActiveRecord::Migration
+  def change
+    create_table :provision_derivations do |t|
+      t.references :order_item, index: true
+      t.text :name
+      t.text :value
+
+      t.timestamps
+    end
+    add_index :provision_derivations, :id
```
stackus: An index is automatically created for the primary key. Don't have to duplicate it here.

app/controllers/order_items_controller.rb
```diff
@@ -78,6 +72,13 @@ def stop_service
 
   def provision_update
     authorize @order_item
+
+    ignore = %w(uuid order_item provision_status)
```
stackus: You may want to use `accepts_nested_attributes_for` for this new code block. It looks like you're accepting nested model data or maybe you should be.

Also build does not create a record in the database, it's an alias for `new`.

db/schema.rb
```diff
@@ -11,11 +11,12 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150225203330) do
+ActiveRecord::Schema.define(version: 20150303012758) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
   enable_extension "uuid-ossp"
+  enable_extension "tablefunc"
```
stackus: Still not seeing a migration for this. Also are you merging rows into column data somewhere? Not sure why this would be added at all.

OBCENEIKON: This actually was meant to be removed, I was originally trying to use a pivot table for the provision_derivations table.

app/controllers/order_items_controller.rb
OBCENEIKON: This was also meant to be removed, I originally had an id in there, not realizing one would be automatically created.

db/migrate/20150227033611_create_provision_derivations.rb
OBCENEIKON: This was also meant to be removed, I originally had created an id not realizing it was automatically created.

PR #309, opened 2015-03-04
Readded password and status_msg to order_items table.

OBCENEIKON:


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/feature/JEL-277-Update_order_items_table">analyzed this pull request</a>.

stackus:
I don't see any migrations tied to these changes.

chriskacerguis:
What's password for?

jntullo:
When creating an RDS instance, the password is never passed back once you set it. So, we need a place to store it.

chriskacerguis:
@jntullo Wouldn't that be stored in the "provider_response" JSON column?

chriskacerguis:
@OBCENEIKON aren't we getting rid of the order_derivations table?  Looks like it's still here?

jntullo:
@chriskacerguis no, it doesn't get passed back in the response, it's a write-only property

PR #316, opened 2015-03-04
Feature/jel 149 add archived flag to projects

OBCENEIKON:
I must of forgotten to create a pull request for this item.


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/feature/JEL-149-Add_archived_flag_to_projects">analyzed this pull request</a>.

chriskacerguis:
@OBCENEIKON looks like we can't auto-merge.  Can you please take a look.

app/models/project.rb
```diff
@@ -49,6 +51,8 @@ class Project < ActiveRecord::Base
 
   # Scopes
   scope :main_inclusions, -> { includes(:staff).includes(:project_answers).includes(:services) }
+  scope :active, -> { where(archived: nil) }
+  scope :archived, -> { where.not(archived: nil) }
```
brittballard: Should we add a spec for these two scopes?

chriskacerguis: @OBCENEIKON ??

brittballard: I think it would be a pretty straight forward spec. I'd say just add two projects one that matches the condition and one that doesn't and then just call the scope and ensure you get the right result back. I can pair with you on it if it would help.

PR #323, opened 2015-03-05
Add demo unit test for AwsFog class

brittballard:


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/bb-tests">analyzed this pull request</a>.

app/workers/provision_worker.rb
```diff
@@ -4,7 +4,7 @@ def initialize(order_item_id)
   end
 
   def perform
-    if miq_settings[:enabled]
+    if miq_settings[:enabled] == 't'
```
stackus: Assuming this is coming from settings_hash any `:enabled` setting is converted into a boolean, all others are left as strings.

brittballard: For some reason it didn't appear to be working. I didn't look into why though.

stackus: Is there a place that is putting 't' in instead of 'true'? That will break UX since it expects 'true' and 'false' values for checkboxes, so where ever that's happening it needs to get fixed.

app/workers/aws_fog.rb
```diff
@@ -4,7 +4,6 @@ def initialize(order_item_id)
   end
 
   def provision
-    mock_mode
```
jntullo: @brittballard curious as to the removal of this. Will we be moving away from mock_mode for demos?

app/workers/provision_worker.rb
brittballard: I'm setting it here: https://github.com/projectjellyfish/api/blob/bb-tests/spec/support/aws_settings_helpers.rb#L8

It looks like maybe i need to use a string instead of the boolean. I think Postgres outputs booleans as "t" and "f". I'll change the setup and remove the `==`.

brittballard: Hmm that doesn't seem to work either. I'll poke around.

app/workers/provisioner.rb
```diff
@@ -1,12 +1,10 @@
 class Provisioner < Providers
+  private
+
   def order_item
     @order_item ||= OrderItem.find @order_item_id
   end
 
-  def mock_mode
```
chriskacerguis: @brittballard what's happening with mock_mode we need that

app/workers/provision_worker.rb
jntullo: @brittballard is it just not getting any value at all? I wasn't having any issues with this earlier when I tested.

brittballard: Oh I need to set the field_type it looks like.

app/workers/provisioner.rb
```diff
@@ -1,12 +1,10 @@
...
-    ENV['MOCK_MODE'] == 'true' ? Fog.mock! : Fog.unmock!
-  end
```
chriskacerguis: @jntullo Wondering if we should tie "MOCK_MODE" into the RAILS_ENV (dev and test = mock mode on, production = mock mode off)

app/workers/provision_worker.rb
brittballard: Alright that fixed it.

app/workers/aws_fog.rb
brittballard: I think we should add a rake task to turn on mocking for demos, or setup a unique environment. Having that in production code makes me nervous.

app/workers/provisioner.rb
brittballard: I think we should come up with a way to set it to on outside of the code itself. Either in the environment setup, or provide a rake task to run on the environment before a demo.

chriskacerguis: From below:

Wondering if we should tie "MOCK_MODE" into the RAILS_ENV (dev and test = mock mode on, production = mock mode off)  

@brittballard thoughts?

app/workers/provisioner.rb
brittballard: If we take this route we should put it in the specific environment rb files I think. Branching on environment is a code smell, but sometimes is necessary.

brittballard: Or we could do an initializer.

chriskacerguis: @brittballard I'm ok with that.  Could you make that change?

chriskacerguis: I like the idea of the environment rb files

brittballard: Alright, initializer added.

app/workers/aws_fog.rb
jntullo: @brittballard ok that makes sense

app/workers/provisioner.rb
chriskacerguis: cool

brittballard: Just saw your environment file comment, @calebthompson pointed out that since this is gem setup it makes more sense as an initializer. That's why I went that route.

chriskacerguis: Initialzier is great.

PR #331, opened 2015-03-09
Feature/jel 148 basic cms and api

OBCENEIKON:
I must of forgotten to create a pull request for this item.


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/feature/JEL-148-Basic_CMS_and_API">analyzed this pull request</a>.

stackus:
You may want to consider looking into a 'versioning' gem such as https://github.com/airblade/paper_trail.

app/controllers/content_pages_controller.rb
```diff
@@ -0,0 +1,99 @@
+class ContentPagesController < ApplicationController
```
stackus: Code Climate is complaining about complex code in this controller. Suggest the following.

- Apply similar refactoring that was done to StaffController, ProjectsController.
- Remove the versioning logic in favor of a gem (suggested paper_trail)

app/models/content_page.rb
```diff
@@ -0,0 +1,26 @@
...
+#  index_content_pages_on_slug      (slug) UNIQUE
+#  index_content_pages_on_staff_id  (staff_id)
+#
+
+class ContentPage < ActiveRecord::Base
+  acts_as_paranoid
+
+  has_many :content_page_revisions, foreign_key: 'content_pages_id'
+  attr_accessor :content_page_revisions_attributes
+  accepts_nested_attributes_for :content_page_revisions
```
stackus: Probably shouldn't expect `create` or `update` to submit revision data.

app/models/content_page.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+#  index_content_pages_on_slug      (slug) UNIQUE
+#  index_content_pages_on_staff_id  (staff_id)
+#
+
+class ContentPage < ActiveRecord::Base
+  acts_as_paranoid
+
+  has_many :content_page_revisions, foreign_key: 'content_pages_id'
+  attr_accessor :content_page_revisions_attributes
```
stackus: This is unnecessary.

app/models/content_page_revision.rb
```diff
@@ -0,0 +1,21 @@
...
+#  body             :text
+#
+# Indexes
+#
+#  index_content_page_revisions_on_content_pages_id  (content_pages_id)
+#  index_content_page_revisions_on_staff_id          (staff_id)
+#
+
+class ContentPageRevision < ActiveRecord::Base
+  belongs_to :content_page
```
stackus: You're including a `staff_id` but don't have a `belongs_to :staff` relationship defined here.

app/policies/content_page_policy.rb
```diff
@@ -0,0 +1,37 @@
+class ContentPagePolicy < ApplicationPolicy
+  def index?
+    true
+  end
+
+  def create?
+    admin_or_related
```
stackus: Each of these `admin_or_related` should either be replaced with `user.admin?` or the contents of `admin_or_related` should be checked if the user it related to the model.

db/migrate/20150305002252_create_content_pages.rb
```diff
@@ -0,0 +1,25 @@
+class CreateContentPages < ActiveRecord::Migration
+  def change
+    create_table :content_pages do |t|
+      t.references :staff, index: true
+
+      t.timestamps
```
stackus: Confirm with @chriskacerguis but I think the standard going forward is to put timestamps first in a table definition.

app/controllers/content_pages_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+    respond_with_params @content_page
+  end
+
+  api :POST, '/content_pages', 'Creates a content page'
+  param :title, String, required: true, desc: 'The title of this content page.'
+  param :body, String, required: false, desc: 'The body of this content page.'
+  error code: 422, desc: ParameterValidation::Messages.missing
+
+  def create
+    @content_page = ContentPage.create @content_page_params
```
calebthompson: Since we're not rendering views, there's no need to use instance variables in the controllers.

app/controllers/content_pages_controller.rb
calebthompson: Specifically,

- refactor away before_actions to load content
- memoize all content helper methods
- authorize within memoized content helper methods
- avoid ivars

app/controllers/content_pages_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  def load_content_page_params
+    params_strip_tags
+    params = ActionController::Parameters.new(@stripped_params)
+    @content_page_params = params.permit(:staff_id, :title, :body)
+  end
+
+  def load_content_page
+    @content_page = ContentPage.find_by_slug!(params.require(:slug))
```
calebthompson: https://github.com/projectjellyfish/api/pull/331/files#r26061354 Especially for things that are returning values - this variable isn't needed at all and we could rename this method `content_page` and call it directly as a value:

```
def destroy
  respond_with content_page.destroy
end

def content_page
  ContentPage.find_by_slug!(params[:slug]).tap { |c| authorize(c) }
end
```

app/controllers/content_pages_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+    authorize @content_page
+    @content_page.destroy
+    respond_with @content_page
+  end
+
+  private
+
+  def load_content_page_params_with_slug
+    params_strip_tags
+    slug = ActionController::Base.helpers.strip_tags(@stripped_params[:title].to_s).parameterize
```
calebthompson: This has already been stripped.

app/controllers/content_pages_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+    content_page_check = ContentPage.where('slug LIKE ?', "#{slug}%")
+    slug = content_page_check.count > 0 ? "#{slug}-#{content_page_check.count + 1}" : slug
```
calebthompson: This is model behavior - which is to say that it should be a part of creation logic on the model or in a builder object that creates the model from params, but it shouldn't be in the controller.

app/controllers/content_pages_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+    @stripped_params[:slug] = slug
+    params = ActionController::Parameters.new(@stripped_params)
```
calebthompson: params_strip_tags should be returning a new ActionController::Parameters, since that logic happens each time it is used.

app/controllers/content_pages_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def load_content_page
+    @content_page = ContentPage.find_by_slug!(params.require(:slug))
+  end
+
+  def load_content_pages
+    @content_pages = query_with ContentPage.all, :pagination
+  end
+
+  def params_strip_tags
```
calebthompson: I would say that this whole thing is too logic-y for a controller, and should be replaced by a service object that:

- Accepts the params from the controller
- Sanitizes them (strips the tags)
- Builds the ContentPage
- Ensures the uniqueness of the slug

app/models/content_page.rb
```diff
@@ -0,0 +1,26 @@
...
+# Indexes
+#
+#  index_content_pages_on_slug      (slug) UNIQUE
+#  index_content_pages_on_staff_id  (staff_id)
+#
+
+class ContentPage < ActiveRecord::Base
+  acts_as_paranoid
+
+  has_many :content_page_revisions, foreign_key: 'content_pages_id'
```
calebthompson: Could this just be called revisions? No reason to "namespace" the model since we don't have another revision class.

app/controllers/content_pages_controller.rb
stackus: Slugs, and logic for them and the record could be offloaded to https://github.com/norman/friendly_id

db/migrate/20150305002252_create_content_pages.rb
chriskacerguis: Yes, timestamps first please.

brittballard:
Just out of curiosity, what is the plan for content pages?

chriskacerguis:
@brittballard to allow users to add arbitrary pages of just info (e.g. procedure to get help, or how to create a project, etc).

brittballard:
@chriskacerguis that's what I suspected. Thank you.

spec/requests/content_pages_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      expect(response.status).to eq(204)
+    end
+
+    it 'returns an error when the content page does not exist' do
+      delete '/content_pages/this-slug-does-not-exist'
+      expect(response.status).to eq(404)
+      expect(JSON(response.body)).to eq('error' => 'Not found.')
+    end
+  end
+end
```
brittballard: Nice job on the specs @OBCENEIKON !

app/controllers/content_pages_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  api :POST, '/content_pages', 'Creates a content page'
+  param :title, String, required: true, desc: 'The title of this content page.'
+  document_content_page_params
+  error code: 422, desc: ParameterValidation::Messages.missing
+
+  def create
+    page = ContentPage.create content_page_params
+    authorize page
```
calebthompson: It's not super useful to authorize after the fact - we probably want to authorize `ContentPage` first, then create the new page.

spec/requests/content_pages_spec.rb
calebthompson: !! Awesome

app/controllers/content_pages_controller.rb
OBCENEIKON: I was wondering about that, I saw the same thing being done in the staff_controller.rb
```
  def create
    staff = Staff.new staff_params
    authorize staff
    staff.save
    respond_with staff
  end
```

stackus: This won't persist the record until the save is hit because it uses `.new`. `.create` will persist immediately.

calebthompson: Yeah, so it's a subtle difference. Staff.create(x) is equivalent to s = Staff.new(x); s.save, but only in exactly that order. Combining the two means that we persist the record then check whether we were allowed to.

Although if authorize rolls back the transaction, we might still be okay.

Issue #128, opened 2015-01-26
Create pre/post hooks in Jellyfish


chriskacerguis:
Here's the list of events that I see (there should be a BEFORE and an AFTER hook)

- [x] User login
- [x] User logout
- [x] Create project
- [ ] Archive project
- [x] Edit project
- [x] Delete project
- [x] Project approval
- [x] Project rejection
- [x] User creation
- [x] User deletion
- [x] User added to project 
- [ ] User removed from project
- [ ] OK Alert created
- [ ] WARNING Alert created
- [ ] CRITICAL Alert created
- [ ] Alert transition (going from one status to another)
- [ ] Add to Cart
- [ ] Remove from Cart
- [ ] Provision
- [x] Add product
- [ ] Edit product
- [ ] Make product active
- [ ] Make product inactive

mafernando:
Looked at a couple different approaches to addressing this issue today.

Current approach is to break down this task into two subtasks:
1) Create a pre/post hook system which could/should tie into the module system being worked on by Thoughtbot
2) Implement a notification system using delayed jobs to avoid holding up the event in question by waiting for the notification to resolve

Am avoiding using pre/post commit hooks defined on models because that would cause the hooks to be invoked whenever we manipulate objects which may not always be desirable. One example would when manipulating objects from within the console.

chriskacerguis:
@mafernando what events are we going to provide hooks for?

mafernando:
@chriskacerguis, I'm working off the list you provided initially and adding onto them as use cases come to mind. I've removed the hooks gem as it looks like that we can get what we want just using plain old ruby class methods. Am working out the details between how the inter-play between the hooks and notification system should look like. 

Here is the current list I'm coding against:

Admin = Admin Only
User = Admin AND User

List of events (work in progress):
- [ ] User: Created new project
- [ ] Admin: Need to approve project
- [ ] User: Project Approved / Rejected
- [ ] User: New Alert
- [ ] User: Alert over
- [ ] User: Service X provisioned
- [ ] User: Checkout Summary

Examples of different notification types:
- Email (Text or HTML)
- SMS 
- Push
- Slack
- Custom

chriskacerguis:
So, here's the list of events that I see (there should be a BEFORE and an AFTER hook)

- [ ] Create project
- [ ] Archive project
- [ ] Edit project
- [ ] Delete project
- [ ] Project approval
- [ ] Project rejection
- [ ] User creation
- [ ] User deletion
[...]

PR #336, opened 2015-03-09
Feature/add hooks system

mafernando:
Addresses Issue #128.

Adds hooks to the following controllers
- Project_approvals
- Project_staff
- Projects
- Sessions
- Staff
- Manage IQ Products
- Products
- Alerts

For an example of how to use these hooks, see [jellyfish-notice](https://github.com/projectjellyfish/jellyfish-notice). To add subscriptions in API, put them in a file under `config/initializers`.

### DIFF
--- a/app/controllers/alerts_controller.rb
+++ b/app/controllers/alerts_controller.rb
@@ -1,6 +1,8 @@
 class AlertsController < ApplicationController
   after_action :verify_authorized
+  after_action :post_hook
 
+  before_action :pre_hook
   before_action :load_alerts, only: [:index]
   before_action :load_alert, only: [:show, :edit, :update, :destroy]
   before_action :load_update_params, only: [:update]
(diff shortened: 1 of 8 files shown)


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/feature/add-hooks-system">analyzed this pull request</a>.

app/controllers/alerts_controller.rb
```diff
@@ -178,4 +180,12 @@ def last_alert_for_service
     }
     Alert.where(conditions).order('updated_at DESC').limit(1).first
   end
+
+  def pre_hook
```
stackus: Something to consider. You can define these two methods in ApplicationController so you don't have to repeat them in every controller you want to apply hooks.

mafernando: @stackus - I only attached hooks to the the controllers associated with the list of events in #128. 
I didn't want all the controllers to start instrumenting events unnecessarily. 
If that isn't an issue, then I agree that defining them on the application controller makes sense. Futhermore, I could just overwrite the pre/post_hook methods in the unsubscribed controllers to be stubbed. Am open to your direction here.

stackus: Sorry, I meant only moving the methods to `ApplicationController` not the `before_filter`s. Since the `before_filter`s are just two lines right now I can't make a big case to make them a concern but that is also an option.

PR #343, opened 2015-03-11
Extract Jellyfish::Fog::AWS for fog/aws support

calebthompson:
This is an example of how modules that provide products will look. There's still a little bit of work to do for swapping between the product types (currently hard coded to manageiq), but since none of this was hooked up before that's out of scope.

I'm PR-ing this in api for context, but it'll eventually live in its own repository.

* Each product type becomes a active record subclass that holds data about its order_item-level configuration (the product_answers and product_questions that were specific to that are represented as columns on this table)
* Each product type row can be related to an order_item as provisional.
* Each product type knows the class that knows how to do actual provisioning, all of which are provisioner subclasses.
* The json included in this PR will be provided by a new class which will map to a subset of current ProductType's current behavior - that it knows how to ask questions that won't ever change. For Fog I don't think there are any such questions, but ManageIQ has a few. This isn't in scope of this change.


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/jellyfish-fog-aws">analyzed this pull request</a>.

calebthompson:
This shouldn't be merged except by me, so I can make sure that nothing is breaking.

app/workers/provisioner.rb
```diff
@@ -1,63 +1,36 @@
...
-    order_item.payload_response = object.to_json
-  end
+  private
 
-  def save_request(request)
-    order_item.payload_request = request.to_json
+  def self.perform
+    yield OrderItem.find(order_item_id)
+  rescue => e
+    warning_retirement_error(order_item, e.message)
```
stackus: Is this accidental or is the name confusing?

app/workers/provisioner.rb
```diff
@@ -1,63 +1,36 @@
...
-  def order_item_details
-    details = {}
-    answers = order_item.product.answers
-    order_item.product.product_type.questions.each do |question|
-      answer = answers.select { |row| row.product_type_question_id == question.id }.first
-      details[question.manageiq_key] = answer.nil? ? question.default : answer.answer
-    end
-    details
+  def self.provision(order_item_id)
+    perform { |order_item| new(order_item).provision }
```
stackus: It's not clicking ATM why you're using a block for order_item lookup. Can you explain?

doc/big_data.json
```diff
@@ -0,0 +1,23 @@
+{
```
stackus: This additional file seems out of place? It might not, can you explain?

lib/jellyfish_fog_aws/lib/jellyfish_fog_aws/databases.rb
```diff
@@ -0,0 +1,50 @@
...
+    def provision
+      db_instance_id = "id-#{order_item.uuid[0..9]}"
+      begin
+        connection.create_db_instance(db_instance_id, details)
+      rescue Excon::Errors::BadRequest, Excon::Errors::Forbidden => e
+        raise e, 'Bad request. Check for valid credentials and proper permissions.', e.backtrace
+      end
+
+      order_item.instance_name = db_instance_id
+      order_item.password = BCrypt::Password.create(@sec_pw)
```
stackus: `@sec_pw` does not appear to be defined anywhere.

lib/jellyfish_fog_aws/lib/jellyfish_fog_aws/databases.rb
```diff
@@ -0,0 +1,50 @@
...
+        aws_secret_access_key: aws_settings[:secret_key]
+      )
+    end
+
+    def identifier
+      order_item.payload_response['data']['body']['CreateDBInstanceResult']['DBInstance']['DBInstanceIdentifier']
+    end
+
+    def snapshot
+      "snapshot-#{order_item.uuid[0..5]}"
```
stackus: Should either of these methods use memoization?

`connection`, `snapshot`

lib/jellyfish_fog_aws/lib/jellyfish_fog_aws/databases.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def connection
+      Fog::AWS::RDS.new(
+        aws_access_key_id: aws_settings[:access_key],
+        aws_secret_access_key: aws_settings[:secret_key]
+      )
+    end
+
+    def identifier
+      order_item.payload_response['data']['body']['CreateDBInstanceResult']['DBInstance']['DBInstanceIdentifier']
```
stackus: If the structure changes getting to the identifier will fail. Gut tells me we should be parsing the response to get to values we feel are important and storing those along side the response.

The alternative is to react to AWS/Fog API response changes and include a version check here to support older data stored everywhere.

lib/jellyfish_fog_aws/lib/jellyfish_fog_aws/infrastructure.rb
```diff
@@ -0,0 +1,41 @@
...
+    def connection
+      Fog::Compute.new(
+        provider: 'AWS',
+        aws_access_key_id: aws_settings[:access_key],
+        aws_secret_access_key: aws_settings[:secret_key]
+      )
+    end
+
+    def server_identifier
+      order_item.payload_response['id']
```
stackus: A whole lot less scary but still the same 'parse the response' comment.

lib/jellyfish_fog_aws/lib/jellyfish_fog_aws/storage.rb
```diff
@@ -0,0 +1,36 @@
...
+    def connection
+      Fog::Storage.new(
+        provider: 'AWS',
+        aws_access_key_id: aws_settings[:access_key],
+        aws_secret_access_key: aws_settings[:secret_key]
+      )
+    end
+
+    def storage_key
+      order_item.payload_response['key']
```
stackus: A whole lot less scary but still the same 'parse the response' comment.

lib/jellyfish_fog_aws/spec/dummy/README.rdoc
```diff
@@ -0,0 +1,28 @@
+== README
```
stackus: Is dummy as a full Rails app needed?

lib/jellyfish_fog_aws/spec/dummy/dummy/README.rdoc
```diff
@@ -0,0 +1,28 @@
+== README
```
stackus: Officially not following now.

dummy/dummy?

app/workers/provisioner.rb
calebthompson: I don't follow your question.

doc/big_data.json
calebthompson: They're not used now, but these are the data that will be required at order time for these products. We can use these to generate the forms:

lib/jellyfish_fog_aws/lib/jellyfish_fog_aws/databases.rb
calebthompson: I like the idea of storing relevant info in addition to the raw payload.

lib/jellyfish_fog_aws/spec/dummy/README.rdoc
calebthompson: It's just what the engine generated. We're not using all of it but it also doesn't hurt us much.

lib/jellyfish_fog_aws/spec/dummy/dummy/README.rdoc
calebthompson: OO this is a mistake.

calebthompson: Do you not like my triply nested Rails app?

doc/big_data.json
calebthompson: Oh and this one specifically was just the only product type that didn't map to something I was pulling out for Fog.

app/workers/provisioner.rb
stackus: For the `perform`, but I think it was `save_request` when I made my comment, method calls a method `warning_retirement_error` to record the error. It seemed like an error to use the 'retirement' error vs a 'provision' error. I wasn't fully sure if it was an error to use that method or if the name was confusing to me because it included 'retirement'.

calebthompson: Oh huh. This method shouldn't even be here anymore.

calebthompson: Oh no I see what you're saying. Something got lost when this block was introduced.

stackus:
:+1: No issues I see that can't be addressed during the process of pulling it out into a gem.

app/models/order_item.rb
```diff
@@ -48,7 +48,7 @@ class OrderItem < ActiveRecord::Base
 
   # Hooks
   before_create :inherit_price_data
-  after_commit :provision, on: :create
+  after_create :provision
```
brittballard: I think if you rebase master in you can switch this back if there is any reason to do so.

app/models/order_item.rb
```diff
@@ -57,6 +57,16 @@ class OrderItem < ActiveRecord::Base
   # Columns
   enum provision_status: { ok: 0, warning: 1, critical: 2, unknown: 3, pending: 4, retired: 5 }
 
+  delegate :provisioner, to: :product
+
+  def answers
```
brittballard: I'm surprised to see this in the diff. Did you actually add this?

calebthompson: Yup

app/models/order_item.rb
calebthompson: What's the flag to not use transactions?

lib/jellyfish_fog_aws/db/migrate/20150309181621_create_jellyfish_fog_aws_infrastructure_products.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJellyfishFogAwsInfrastructureProducts < ActiveRecord::Migration
+  def change
+    create_table :jellyfish_fog_aws_infrastructure_products do |t|
+      t.string :instance_size
```
brittballard: Why you no timestamp?

doc/big_data.json
brittballard: Did you mean to put this in doc? If so ignore me.

lib/jellyfish_fog_aws/db/migrate/20150309181621_create_jellyfish_fog_aws_infrastructure_products.rb
calebthompson: Ugh I forgot about that.

lib/jellyfish_fog_aws/spec/dummy/README.rdoc
brittballard: @stackus is there a way to get the engine to generate less? I had the same thought when working with @mafernando 's gem. It annoyed me we had to add sqlite3 gem as a dev dependency when we weren't using the database.

app/models/order_item.rb
brittballard: `type: :integration`

doc/big_data.json
calebthompson: Stack and I talked about this. It doesn't make sense to us in doc but we don't have a better place. I might translate it to rubby or yaml and leave it in the models. We'll see.

calebthompson: But moving it, I think, is out of scope.

brittballard: Cool, makes sense to me.

PR #344, opened 2015-03-12
Bug/jel 147 dissallow checkout if budget spent

OBCENEIKON:


chriskacerguis:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/54d4fa64695680762a0024c5/compare/bug/JEL-147-Dissallow_checkout_if_budget_spent">analyzed this pull request</a>.

app/models/order.rb
```diff
@@ -45,11 +47,21 @@ def item_count_for_project_id(pid)
   def total_per_order(oid)
```
stackus: If the `calculate_price` method was on `OrderItem` (and used properties instead of arguments) then this method could become.

```
def total_per_order(order_id)
  order_items.where(order_id: order_id).map(&:calculate_price).sum
end
```

`hours_in_month` would need to be defined as a constant somewhere or in `OrderItem` or even left as an optional parameter to the method.

app/models/order.rb
```diff
@@ -45,11 +47,21 @@ def item_count_for_project_id(pid)
...
-      total = total + order_item.setup_price + order_item.monthly_price + (order_item.hourly_price * 750)
+      total += calculate_price(order_item.setup_price, order_item.hourly_price, order_item.monthly_price)
     end
     total
   end
 
+  def exceeds_budget?
+    grouped_order_items = order_items.group_by(&:project)
+    grouped_order_items.reduce(false) do |over_budget, project_order_items|
+      cost = project_order_items[1].reduce(0) do |sum, oi|
```
stackus: Doing the previous suggestion and also applying it here should help with the CodeClimate 'Complex method' issue.

OBCENEIKON: Thanks @stackus I was wanting to do that from the beginning but couldn't figure out how.

app/models/order.rb
```diff
@@ -43,11 +45,15 @@ def item_count_for_project_id(pid)
...
-    total = 0
-    order_items.where(order_id: oid).each do |order_item|
-      total = total + order_item.setup_price + order_item.monthly_price + (order_item.hourly_price * 750)
+    order_items.where(order_id: oid).map(&:calculate_price).sum
+  end
+
+  def exceeds_budget?
+    grouped_order_items = order_items.group_by(&:project)
+    grouped_order_items.reduce(false) do |over_budget, project_order_items|
+      cost = project_order_items[1].map(&:calculate_price).sum
```
stackus: Do you really want the second item in the results list? You may mean `[0]` here or even better `.first`.

OBCENEIKON: ```[0]``` is actually the project, whereas ```[1]``` is the order item.

OBCENEIKON: This comment did point out an error to me though :)
Forgot to remove reference to the concern file that was removed.

stackus: Yeap, you're right. `reduce` on a hash will give you `[key, values]` and in this case that's `[<Project>, [<OrderItem>, ...]]`.

You will want to add `.includes(:project)` just before `.group_by` to squish a N+1 query issue though.

OBCENEIKON: Awesome, I wouldn't have known about that. Thanks!

PR #351, opened 2015-03-16
Fix merge issues with jellyfish_fog_aws

calebthompson:
* Re-add missing Infrastructure spec
* Add missing data from database products table

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -19,21 +19,23 @@ class Setting < ActiveRecord::Base
     hid
   end
 
-  def nil_values(field)
-    field['value_withheld'] && field['value'].nil?
-  end
-
   # Returns a hash of setting_fields with their hids as the key and their value as the value
   # check_box fields are considered booleans and anything but 'true' is false
   def settings_hash
     setting_fields.map do |setting|
       value = case setting.field_type
               when 'check_box'
-                'true' == setting.value ? true : false
+                'true' == setting.value
               else
                 setting.value
               end
       [setting.hid.to_sym, value]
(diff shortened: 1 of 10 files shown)


app/models/setting.rb
```diff
@@ -19,21 +19,23 @@ def to_param
...
-  end
-
   # Returns a hash of setting_fields with their hids as the key and their value as the value
   # check_box fields are considered booleans and anything but 'true' is false
   def settings_hash
     setting_fields.map do |setting|
       value = case setting.field_type
               when 'check_box'
-                'true' == setting.value ? true : false
+                'true' == setting.value
```
stackus: Why is the transform to a boolean being removed?

calebthompson: It's not. `#==` returns a boolean here, so the ternary was unnecessary.

stackus: :facepalm:

Sorry, so it is. I only saw the red removed bits.

lib/jellyfish_fog_aws/spec/dummy/app/models/provisioner.rb
```diff
@@ -1 +1,9 @@
-class Provisioner; end
+class Provisioner
+  attr_reader :order_item
+
+  def initialize(order_item)
+    @order_item = order_item
+  end
+
+  def aws_settings; end
```
brittballard: I'm surprised this is the preferred style. I've always undone this when I stumble across it, I guess I should stop that.

brittballard:
Pretty straight forward. LGTM.

PR #352, opened 2015-03-17
Port AwsFog specs over to the new gem

brittballard:


chriskacerguis:
@brittballard looks like some fails here, and I can't auto-merge.  Could you take a look?

brittballard:
Interesting, sure.

brittballard:
All fixed, that was an annoying little bug I introduced there.

PR #354, opened 2015-03-18
Combine API + UX projects

calebthompson:
Launching the app is now accomplished by running `foreman start`, which
starts each of node, puma, and rake jobs:work processes.


calebthompson:
This is the only commit (so far) that I actually added: https://github.com/projectjellyfish/api/commit/ee5e16c5793515ff2d5f7dc6e266af2c57f9e6b0

The rest is the result of a subtree merge: `git subtree add --prefix=app/assets/javascripts/ux git@github.com:projectjellyfish/ux master`

calebthompson:
The subtree merge allowed us to keep the history (logs are a little broken because the subtree merge commit moved all the files).

There's another approach I could take, it would require rewriting the entire history of ux using `git-filter-branch`. http://stackoverflow.com/a/1425914/218211

stackus:
Going to have to download and vet.

Procfile
```diff
@@ -1 +1,3 @@
-web: bundle exec puma -C config/puma.rb
+web: bundle exec puma -C config/puma.rb --port 3000
```
calebthompson: Whoops. This is causing puma to listen on two ports, which means that the hard-coded value I can't find somewhere in ux that points to localhost:3000 doesn't break. Need to fix it.

Issue #361, opened 2015-03-23
Return mobile_token on successful sign in


mafernando:
Requirements:
- [x] Return mobile_token on successful sign in
- [x] Email Chris upon completion

mafernando:
Reopening per problem direction from @chriskacerguis

mafernando:
@chriskacerguis - we need to have API check against a user's encrypted token when they attempt to auth with token vs using the hard coded token value currently in place.

chriskacerguis:
Ok

jklarfeld:
Additionally it does not seem that a new token can be requested at any time.  A POST to the /staff/sign_in results in a token the first time, but a second POST is a 500 error.

I also cannot find any method of obtaining a new token using only the existing token (this would be massively helpful to prevent having to store a password on the device).

mafernando:
@jklarfeld - you should now be able to login to API with 
```
POST /api/v1/staff/sign_in.json HTTP/1.1
Content-Type: application/json
Host: localhost:3000
Connection: close
User-Agent: Paw/2.1.1 (Macintosh; OS X/10.9.5) GCDHTTPRequest
Content-Length: 71

{"staff":{"email":"email","password":"password"}}
```

and get back the following json

```
{
  "id": 1,
  "first_name": "first",
  "last_name": "last",
  "email": "email",
  "phone": null,
  "role": "role",
  "created_at": "2015-03-27T01:21:50.791Z",
  "updated_at": "2015-03-27T01:25:30.818Z",
  "api_token": "api_token"
}
```

subsequent API requests would then provide the user email and api_token from sign in

```
GET /api/v1/products HTTP/1.1
X-Staff-Token: api_token
X-Staff-Email: email
Content-Type: application/x-www-form-urlencoded
Host: localhost:3000
Connection: close
User-Agent: Paw/2.1.1 (Macintosh; OS X/10.9.5) GCDHTTPRequest

```

Let me know if you encounter any issues here.

mafernando:
Thanks to @calebthompson for helping me work through this one. 

Based on our conversation, we could use the [devise token auth gem](https://github.com/lynndylanhurley/devise_token_auth) to add more functionality to api token auth. And we have some great [devise token auth information](https://github.com/plataformatec/devise/wiki/How-To:-Simple-Token-Authentication-Example) we can refer to for this.

[...]

PR #364, opened 2015-03-24
Adds mobile_token to JSON sign in response if successful.

mafernando:
- Added lines to sessions controller that generate and persist a new token when user signs in via JSON.
- Added mobile_token attribute accessor to Staff model.
- Added mobile_token attribute to StaffSerializer.

Address issue #361

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -21,7 +21,7 @@ class SessionsController < Devise::SessionsController
 
         if resource
           sign_in(resource_name, resource)
-          resource.mobile_token = resource.secret = SecureRandom.hex
+          resource.api_token = resource.secret = SecureRandom.hex
           resource.save # GENERATE NEW TOKEN FOR USER AND PERSIST IT TO DB
           render json: resource
         else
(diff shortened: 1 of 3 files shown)


mafernando:
Small note - if you are trying to sign in to API via JSON, you'll need to add a `.json` to the end of the sign in URL for it to be identified as JSON in API.
```
http://localhost:3000/staff/sign_in.json
```
Additionally, the staff email and password need to exist as JSON attributes within a JSON staff object.
```
{"staff":{"email":"admin@projectjellyfish.org","password":"jellyfish"}}
```

app/controllers/sessions_controller.rb
```diff
@@ -21,6 +21,8 @@ def create
 
         if resource
           sign_in(resource_name, resource)
+          resource.mobile_token = resource.secret = SecureRandom.hex
```
stackus: Didn't like `api_token`?

mafernando: https://github.com/projectjellyfish/api/issues/361

Chris wanted mobile_token, so I decided to go with that.

Changing it seems simple enough going forward, so I'm not too concerned.

mafernando:
@stackus - Chris went ahead with api_token :+1:

PR #368, opened 2015-03-25
Feature/azure module integration

jntullo:
adding in jellyfiish-azure


chriskacerguis:
Can we do this in the gem?

Issue #363, opened 2015-03-24
Build Cron Job for Modules


OBCENEIKON:
Create a way for modules to run cron jobs using the whenever gem.

OBCENEIKON:
@chriskacerguis The only thing I am finding related to the rufus-scheduler process stopping is when the ruby server is stopped. It appears thats the only way the process can be stopped, so jobs will stop running when the application stops running.

chriskacerguis:
Let's run with rufus.  Please create a process to check to make sure it is running, if it hasn't run in 5 minutes, please create an alert.

PR #369, opened 2015-03-25
Removed the whenever gem in favor of rufus-scheduler to allow jellyfish ...

OBCENEIKON:
...modules schedule tasks.


OBCENEIKON:
Closes #363

PR #439, opened 2015-03-27
Feature/return api token json

mafernando:
Addresses issue #361, Return mobile token on JSON sign in  
Changes:
 - Modify session controller to reset user authentication token on login and return authentication token as api token.
 - Removed call to token? from TokenAuthentication because it was incorrectly comparing tokens with BCrypt.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -21,7 +21,7 @@ class SessionsController < Devise::SessionsController
 
         if resource
           resource.authentication_token = nil # RESET USER TOKEN AFTER GETTING RESOURCE FROM WARDEN
-          resource.save # THIS FORCES A NEW API TOKEN TO BE GENERATED FOR USER
+          resource.save # THIS FORCES A NEW USER API TOKEN TO BE GENERATED
           sign_in(resource_name, resource)
           resource.api_token = resource.authentication_token
           render json: resource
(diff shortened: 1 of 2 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -20,9 +20,10 @@ def create
         self.resource = warden.authenticate(auth_options)
 
         if resource
+          resource.authentication_token = nil # RESET USER TOKEN AFTER GETTING RESOURCE FROM WARDEN
+          resource.save # THIS FORCES A NEW USER API TOKEN TO BE GENERATED
           sign_in(resource_name, resource)
-          resource.api_token = resource.secret = SecureRandom.hex
-          resource.save # GENERATE NEW TOKEN FOR USER AND PERSIST IT TO DB
+          resource.api_token = resource.authentication_token
```
stackus: By sending the encrypted string you defeat the purpose of encrypting it. Did you mean to send secret instead? I see in the token concern file you're bypassing the token validation and using the token in a lookup.

Another issue
- The token string is not web safe. Which means the characters it uses may not come across a request without issue.

mafernando: @stackus - I've incorporated your feedback into PR https://github.com/projectjellyfish/api/pull/440, let me know if you have any other guidance here. thank you for catching this.

Issue #365, opened 2015-03-25
Update ManageIQ worker OpenSSL::SSL::VERIFY_NONE


OBCENEIKON:
I found a fix for OpenSSL::SSL:VERIFY_NONE

https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview

>On OS/X, for development only, it may be easiest just to disable certificate verification because the certificates are stored in the keychain, not the file system:

```
require "omniauth-facebook"
OpenSSL::SSL::VERIFY_PEER = OpenSSL::SSL::VERIFY_NONE if Rails.env.development? 
config.omniauth :facebook, "APP_ID", "APP_SECRET"
```

PR #442, opened 2015-03-27
Bug/jel 365 update manage iq worker

OBCENEIKON:


OBCENEIKON:
Closes #365

app/workers/manage_iq.rb
```diff
@@ -75,12 +75,15 @@ def payload
   end
 
   def request
-    # TODO: verify_ssl needs to be changed, this is the only way I could get it to work in development.
+    # On OS/X, for development only, it may be easiest just to disable certificate verification because the certificates are stored in the keychain, not the file system
+    # https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview
+    ssl_verify = Rails.env.development? ? OpenSSL::SSL::VERIFY_NONE : OpenSSL::SSL::VERIFY_PEER
```
chriskacerguis: We wouldn't want to do this in production, as that is a security issue.  Let's make sure we only do this in env = "development"

chriskacerguis: @OBCENEIKON ???

calebthompson: That's what this line does.

calebthompson:
LGTM

app/workers/manage_iq.rb
chriskacerguis: /facepalm

PR #445, opened 2015-03-30
Feature/jel 329 modularize login system

### LINKED ISSUE #329: Move "internal login" system to be a module
Move "internal login" system to be a module using pre/post hooks.

OBCENEIKON:
Closes #329


spec/factories/authentications.rb
```diff
@@ -0,0 +1,23 @@
...
+# Indexes
+#
+#  index_authentications_on_provider_and_uid  (provider,uid)
+#  index_authentications_on_staff_id          (staff_id)
+#
+
+FactoryGirl.define do
+  factory :authentications do
+    provider 'facebook'
+    uid '1234567'
```
chriskacerguis: Uhhh...should that "uid" be there???

chriskacerguis: @OBCENEIKON ???

Gemfile
```diff
@@ -36,6 +36,8 @@ gem 'responders'
 # Authentication
 gem 'devise'
 gem 'ruby-saml'
+gem 'omniauth'
+gem 'omniauth-ldap'
```
calebthompson: Do we want this built in? Should this instead be a module we can add to show off how others could do this, as we did with jellyfish_fog_aws and jellyfish-azure?

chriskacerguis: @OBCENEIKON is in the process of moving this to a module.

spec/factories/authentications.rb
OBCENEIKON: Yes, this is for the authentications table. OmniAuth gives back a provider and uid for each provider.

app/models/staff.rb
```diff
@@ -62,4 +63,20 @@ class Staff < ActiveRecord::Base
   def gravatar
     '3fc88b95c85e43f157cb1ffd0e37e832'
   end
+
+  def self.find_by_auth(auth_hash)
+    auth_match = Authentications.find_by(provider: auth_hash['proivider'], uid: auth_hash['uid'].to_s)
```
calebthompson: misspelled 'provider' here. Is there a test on this method?

config/initializers/omniauth.rb
```diff
@@ -0,0 +1,13 @@
+Rails.application.config.middleware.use OmniAuth::Builder do
+  # provider :developer unless Rails.env.production?
+  # provider :dropbox, ENV['DROPBOX_ACCESS_KEY'], ENV['DROPBOX_SECRET_KEY'], path_prefix: '/staff/auth'
+  # provider :ldap,
```
calebthompson: We're using LDAP - is any of this configured?

spec/factories/authentications.rb
calebthompson: From what I'm seeing, we'll want a `uid`, but it should be random, and the db should have a uniq index so we don't have multiple users signed in with a single facebook account.

calebthompson: By "random" I mean something like `sequence(:uid)`, not necessarily random, just different between records.

OBCENEIKON: UID comes from the provider, like a member ID.
https://github.com/intridea/omniauth/wiki/Auth-Hash-Schema

config/initializers/omniauth.rb
OBCENEIKON: I was testing with it configured to jumpcloud

calebthompson: jellyfish-ldap probably should configure this (I assume that or similar is what you're building out).

chriskacerguis:
Closing request after talking with @OBCENEIKON

PR #457, opened 2015-04-01
Enhancement/jel 454 migrate bower to rails assets

AllenBW:


AllenBW:
Just remember, run bundle install after this is pulled in...

chriskacerguis:
Question for @AllenBW and @calebthompson 
Why are we adding all the bower_compnents vs doing a bower install?

AllenBW:
doing neither now bower components will be gone now that rails assets will
be pulled down on bundle install.

PR #462, opened 2015-04-01
initial commit, adjusts version mismatch to fix safari errors.

AllenBW:


Gemfile
```diff
@@ -1,4 +1,4 @@
-ruby '2.2.1'
+ruby '2.1.5'
```
chriskacerguis: We *must* be using Ruby 2.2.1

AllenBW: Yes yes, i know, this commit was a mistake and is being addressed
presently... ends up i HAVE to work through my capybara issues prior to
pushing this fix

Issue #455, opened 2015-03-31
Fix Safari errors (formerly FontAwesome errors) 


sonejah21:
Safari browsers throw FontAwesome errors. 
May be result in alignment issues or could be completely unrelated.  See ticket #451.

Why are we not using the "font-awesome-sass" gem in our rails app? Maybe that would help.

sonejah21:
@calebthompson - I assigned this to you because I need your advice in tackling it.

AllenBW:
this may be fixed by https://github.com/projectjellyfish/api/pull/457, do a pull check it out (nonmac so I can't say for sure)

sonejah21:
Awesome, @AllenBW and @calebthompson  -> Font Awesome errors gone. 

Now I am seeing new errors:
- "TypeError: Object is not a function (evaluating 'doneCallback()')........ angular-animate.js:993"
- "Unexpected CSS token : "


![uploads/79eeee8c-9e24-4e6c-9b7d-01f4e4de5e0b/Screen Shot 2015-04-01 at 9.58.42 AM.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F79eeee8c-9e24-4e6c-9b7d-01f4e4de5e0b%2FScreen+Shot+2015-04-01+at+9.58.42+AM.png)

AllenBW:
Okie dokie the culprit is  a mismatch of angular version and angular-animate versions for whatever reason ang v is 1.3.15 and ang-animate is 1.2.28  fix incoming....

sonejah21:
Ahhhh i see! Nice. I'm going to reassign this to you for now, @AllenBW. Thanks for getting to the bottom of it. Let me know when the fix is in and I'll re-test.

calebthompson:
I'm AFK but Mason and I fixed this in the product types branch (a tweak to the stuff in master) so I'll take it back.

calebthompson:
Other errors also fixed.

AllenBW:
Oh good, because I can't get a build to pass till that WebKit-safari gem
installs... Or I bail to ubuntu

Issue #269, opened 2015-02-23
Allow users to be able to specify the value for a productType item.


chriskacerguis:
Create an entry in ProductTypes ("user_fillable") to allow for a user to add data when adding to cart.

calebthompson:
Is there a separate issue for actually using this field, such that this issue is actually just adding a boolean flag to product_type in the database?

brittballard:
Who is working on the UX portion of this story? This will require some pretty heavy collaboration I suspect.

chriskacerguis:
It will be a combo of @AllenBW and @sonejah21

brittballard:
I can start digging into some Angular if necessary. I just don't want to step on toes.

sonejah21:
@chriskacerguis - should this title read "Allow users to be able to specify the value for a productType item"? 
@brittballard - pending Chris K's opinion and subsequent approval, i say you can dig in wherever you like. Angular help is always appreciated. If you start on this please let me know so I can change gears and assist.

chriskacerguis:
fixed typo in title

brittballard:
@sonejah21 I'm curious where you think this should happen? Should this happen when a person add a service to a project, or when they checkout?

brittballard:
Once we complete this story, I don't know that we'll still need product types for anything beyond organization purposes. I think this will allow us to remove the product questions and answers stuff.

brittballard:
I have changed my tune. We will still need product types.

calebthompson:
This is going to become order_item.answers or similar. It's related to the product type changes I'm doing but I may handle it as a follow-up.

calebthompson:
@chriskacerguis I need clarification on this. I'm of the opinion (and I've discussed it with @masonforest and @brittballard who agree) that all of the questions we just put onto Product in #466 should be moved to OrderItem. This means that we have fewer products and each one gets both sensible defaults and can be customized.

We'd also have questions that lived directly on products, such as AWS credentials which wouldn't vary by order but would differ by product.

This would mean expanding the current diagram:

```
  product_type
  ├── product
  │   └── Provisioning answers JSON column in database
  ├── Provisioning questions JSON object in memory
  │
  └── Rails.application.config.product_types configuration stores types
```

To include order items, where each set of questions and answers are completely distinct (not overrides, only answered in one place):

```
  product_type
  ├── product
  │   └── Provisioning answers like credentials
  ├── order item
  │   └── Specific answers like instance size or availability
[...]

PR #466, opened 2015-04-02
Collect provisioning information in UX

### LINKED ISSUE #346: Store pertinent info from payload_response on order_item or *product

calebthompson:
We've replaced ProductTypes, ProductTypeQuestions, ProductAnswers with
json-schema formatted definitions of each type and the information we
need to gather about it. This information is stored in a json column on
Product.

As this information is stored in memory, it's simple for an engine or
the host application to add or remove ProductTypes (now a PORO wrapper
around the config option), necessitating only an app restart for the
changes to take effect.

      product_type
      ├── product
      │   └── Provisioning answers JSON column in database
      ├── Provisioning questions JSON object in memory
      │
      └── Rails.application.config.product_types configuration stores types

As Products are only displayed in the marketplace if their ProductType
(or one with the same name) is present in the Rails config, Products
related to removed ProductTypes will simply not be displayed, even
though they still exist in the database.

* Use Angular Schema Form to display provisioning_answers questions
  depending on selected ProductType
* Fix various javascript and stylesheet issues
* Use poltergeist in feature specs (Capybara Webkit was having issues
  both in installation and use)

Closes #346


app/assets/javascripts/products/products_resource.js
```diff
@@ -5,13 +5,11 @@ var ProductsResource = function($resource, apiResource) {
   return $resource(apiResource('productsById'), {'id': '@id'}, {
     'update': {
       method: 'PUT',
-      //TODO: This is likely temporary. When the product form gets refactored we will probably rework this.
-      url: apiResource('manageIqProductsById')
+      url: apiResource('productsById')
     },
     'save': {
```
stackus: The save entry can be removed.

app/controllers/products_controller.rb
```diff
@@ -5,24 +5,62 @@ class ProductsController < ApplicationController
...
+  param :name, String, desc: 'Product Name', required: true
+  param :product_type, String, desc: 'Product Type', required: true
+  param :provisioning_answers, Hash, desc: 'Provisioning Answers', required: true
+  param :setup_price, PRODUCT_PRICE_REGEX, desc: 'Initial Setup Fee'
+  error code: 422, desc: ParameterValidation::Messages.missing
+
+  def create
+    product = Product.new(product_params)
+    authorize product
+    logger.error params.inspect
```
stackus: Looks like some debugging was left in.

app/assets/javascripts/admin/products/create_product_controller.js
```diff
@@ -3,15 +3,10 @@
...
-  this.productType = _.find(categories, {id: $stateParams.product_type_id});
-
-  this.product.answers = _.map(this.productType.questions, function(question) {
-    return {
-      answer: question.default,
-      product_type_question_id: question.id,
-      question: question
-    };
-  });
+  this.product.img = "products/aws_ec2.png";
```
jntullo: How does this work? Wouldn't the product image change, is this just a default?

spec/features/create_a_product_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    expect(page).to have_content('Product was successfully added.')
+    expect(Product.last.attributes).to include new_product_attributes
+  end
+
+  def selectize(container, option:)
+    find("#{container} .selectize-input").click
+    find("#{container} #{option}").click
+  end
+end
```
jntullo: This spec is awesome.

app/assets/javascripts/admin/products/create_product_controller.js
calebthompson: This is just the default value. We were seeing 404s because the selectize preview element renders the selected image, which was nothing, so `/assets/images` or whatever was a problem.

doc/storage-miq.json
```diff
@@ -0,0 +1,56 @@
...
+        "us-west-2",
+        "EU",
+        "ap-northeast-1",
+        "ap-southeast-1",
+        "ap-southeast-2"
+      ],
+      "default": ""
+    }
+  }
+}
```
jntullo: Why are these json files here and not in config/product_questions?

calebthompson: Good question. These ones aren't being loaded and I'm unsure whether they should be. They're specific to MIQ, which is probably being pulled out, right? If that's the case, then they belong in that engine not in the main app.

jntullo: Oh ok makes sense. @chriskacerguis what will be doing with MIQ and how should we handle this?

calebthompson:
I know there's a lot to look through in here, but as it blocks a few other issues (#455, #269, others probably) I'd like to get it in and we can deal with fallout as necessary.

jntullo:
Totally fine - we can deal with the MIQ stuff later.

PR #467, opened 2015-04-02
Add disable annotate env var

mafernando:
Disable annotate on migrations by default and check for ENABLE_ANNOTATIONS if in development

### DIFF
--- a/lib/tasks/auto_annotate_models.rake
+++ b/lib/tasks/auto_annotate_models.rake
@@ -20,7 +20,7 @@ if Rails.env.development?
       'exclude_fixtures'     => 'false',
       'exclude_factories'    => 'false',
       'ignore_model_sub_dir' => 'true',
-      'skip_on_db_migrate'   => 'true',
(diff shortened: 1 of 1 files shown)


mafernando:
@stackus - i think this is what you had in mind?

stackus:
Could it be made to be transparent for the developers? As it is these changes will require devs who modify the schema to remember to enable this and to also run the rake command up update the annotations.

mafernando:
Sure. Annotate now does not run by default if in "development" unless 'DISABLE_ANNOTATIONS = true' is an ENV variable

lib/tasks/auto_annotate_models.rake
```diff
@@ -30,5 +30,5 @@ if Rails.env.development?
     )
   end
 
-  Annotate.load_tasks
+  Annotate.load_tasks unless ENV.key? 'DISABLE_ANNOTATIONS'
```
stackus: :+1: 

Also yes `key?` vs the deprecated `has_key?`.

PR #468, opened 2015-04-03
Feature/docker integration

jntullo:
Docker setup and necessary docker files.


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ruby:2.2.1
+RUN apt-get update -qq && apt-get install -y build-essential
+RUN apt-get install make
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev
+RUN apt-get install -y libqt4-webkit libqt4-dev xvfb
+RUN apt-get install -y nodejs
```
calebthompson: The best practice for this is to combine them all to one update and install directive, with each package on its own line and sorted. This is because each intermediate image (each RUN, ENV, WORKDIR, etc.) increases the image size, which is important to keep down.

So this would be something like

```
RUN apt-get update -qq && apt-get install -y \
  build-essential \
  libpq-dev \
  make \
  nodejs
```

calebthompson: https://docs.docker.com/articles/dockerfile_best-practices/

jntullo: cool, I'll fix that :) thanks

PR #470, opened 2015-04-03
fixed width and margins to center bell

tangerine007:
422-bell

### DIFF
--- a/app/assets/stylesheets/common/_layout.sass
+++ b/app/assets/stylesheets/common/_layout.sass
@@ -59,6 +59,10 @@ body
       line-height: 17px
       font-size: 14px
       padding: 9px 12px
+      i
+        margin-left: 25px
+        margin-right: 25px
+        width: 14px
 
 .info-boxes
   display: table
--- a/app/assets/stylesheets/common/_layout.sass
+++ b/app/assets/stylesheets/common/_layout.sass
@@ -59,10 +59,8 @@ body
       line-height: 17px
       font-size: 14px
       padding: 9px 12px
-      i
-        margin-left: 25px
-        margin-right: 25px
-        width: 14px
+      &.centerThis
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/common/_layout.sass
```diff
@@ -59,6 +59,10 @@ body
       line-height: 17px
       font-size: 14px
       padding: 9px 12px
+      i
```
stackus: Have you tested this with a wide and narrow browser window? These look like they're specific to a certain window width.

tangerine007: It is fixed pixel width for the column, I tried yesterday and it kept it at 64px no matter the window size. Heroku isn't launching for me atm so I can't pull it up right now.

chriskacerguis: @tangerine007 Heroku is back up, can you take a look here and let me know?

tangerine007: Heroku is down for me again... very frustrating. I am pretty sure this works fine, unless 
```text-align: center``` would be better?

**Comment by calebthompson** [comment]
:+1:

File: public/templates/partials/common/problem_alerts.html [hunk 2]
```diff
@@ -11,7 +11,7 @@
             </thead>
             <tbody>
             <tr ng-repeat="row in alerts">
-                <td ng-switch on="row.status">
+                <td ng-switch on="row.status" class="centerThis">
```
stackus: Use the bootstrap `text-center` alignment class.

http://getbootstrap.com/css/#type-alignment

tangerine007: I will center this bell if it is the last thing I do.

PR #471, opened 2015-04-03
No more overlap, top icon made shorter

tangerine007:
434-overlap


public/templates/partials/common/left_sidebar.html
```diff
@@ -47,7 +47,7 @@
           </a>
           <ul>
             <li class="first: $first">
-              <i></i>
+              <b></b>
```
stackus: Changing the tag here shouldn't need to happen. The issue that I see is that the class attribute on line 49 isn't the angular version. The tag changes here and in the styles file should be reverted and the attribute changed to `ng-class` so that angular will interpolate the `$first` variable.

chriskacerguis: @tangerine007 thoughts?

tangerine007: Have been having issues with my dev environment, working off of heroku now. Anyway, changing to the ng-class did not work for me.
```
<li ng-class="{first: $first}">
              <i></i>
              <a ui-sref="base.authed.admin.products.list" ui-sref-active="active">Products</a></li>
```

stackus: This is because the style for `li b` is matching but not `.first b` or that it is but the first is taking precedence. Expanding the surrounding code and looking further at it I see that `.first i` should be indented under the `li` and be written as `&.first i`. This would compile to the style `li.first i` and will take precedence over the `li i` style. That said the height and offsets for the sprite for `b` could be applied to `i`, swapping out the element tag out isn't the fix.

Opening up Firebug or the developer console and looking at the applied and overridden styles should answer the precedence issues.

tangerine007: I made another commit, I just hardcoded it to be first since it is not repeating anyway. 
```
<li class="ng-class first">
```

stackus:
You've checked in a .project file and Git. conflict markers.

tangerine007:
ng-scope isn't being added for these lines.
```
<li ng-class="{first: $first}">
```
should be...
```
<li ng-class="{first: $first} class="ng-scope first">
```
when first or...
```
<li ng-class="{first: $first} class="ng-scope">
```
when not first. Problem is not with css.

tangerine007:
Ok, I think problem was it wasn't being repeated so the $first didn't make sense. Is this ok to do in angular?
```
<li class="first">
              <i></i>
              <a ui-sref="base.authed.admin.users.list" ui-sref-active="active">Users</a>
</li>

<li>
             <i></i>
              <a ui-sref="base.authed.admin.products.list" ui-sref-active="active">Products</a>
</li>
...etc...
```

tangerine007:
I see that the .project file and gemfile.lock are still here, I did a hard reset of the branch I thought would remove them but didn't... how can I remove these files from the commit history?

chriskacerguis:
You will need to delete them from your branch (git rm FILENAME), commit them, then re-push your branch.

PR #474, opened 2015-04-03
Revert bin files back to their version prior to the provisioning informa...

stackus:
...tion merge.


stackus:
@calebthompson Was the a reason the files were stubbed or will there be consequences reverting them?

calebthompson:
This wasn't an intentional change on my part. @brittballard @masonforest ?

masonforest:
I didn't intend to change them either. Reverting them should be fine.

PR #482, opened 2015-04-06
Modified the instructions for Clone Repo

mafernando:
Modified the instructions for Clone Repo. SSH key won't be in GitHub, so give it the public URL instead.


mafernando:
Also need to modify the install guide to install ruby 2.2.1 with rbenv.

mafernando:
also [need](http://stackoverflow.com/questions/28056595/bundle-not-working-with-rbenv) to run `gem install bundle` outside of api directory

PR #490, opened 2015-04-08
Make jellyfish core work with jellyfish_fog_aws gem

calebthompson:


chriskacerguis:
:+1:

calebthompson:
Some problems with the syntax. fixing them now.

spec/modules/jellyfish_azure_spec.rb
```diff
@@ -1,61 +1,20 @@
 require 'rails_helper'
 
-def find_gem(name)
```
calebthompson: No real changes to this file, just cleaned up a bit and made sure not to define methods on kernel that could interact with other specs.

PR #498, opened 2015-04-09
Bug/489 autoprefixer for css

sonejah21:
Added in the autoprefixer-rails gem to return to flawless alignment throughout the app.


Gemfile
```diff
@@ -48,6 +48,9 @@ gem 'bootstrap-sass'
 gem 'sass-rails'
 gem 'jellyfish_fog_aws'
 
+# Autoprefixer for CSS classes using Can I Use values.
+gem "autoprefixer-rails"
```
stackus: Single quotes here will hush up rubocop.

sonejah21: thanks!

PR #502, opened 2015-04-10
Enhancement/jel 464 saml metadata functionality

OBCENEIKON:


app/controllers/saml_controller.rb
```diff
@@ -21,10 +33,115 @@ def consume
...
+    settings = idp_metadata_parser.parse_remote('http://sso.projectjellyfish.org:8080/openam/saml2/jsp/exportmetadata.jsp?entityid=https://sso.projectjellyfish.org:8443/openam')
+
+    settings.assertion_consumer_service_url = acs_saml_index_url
+    settings.assertion_consumer_logout_service_url = logout_saml_index_url
+    settings.issuer = "#{metadata_saml_index_url}.xml"
+
+    settings.name_identifier_format = 'urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress'
+    settings.authn_context = 'urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport'
+
+    settings.certificate = '-----BEGIN CERTIFICATE-----
```
mafernando: @chriskacerguis and @OBCENEIKON - should this be in the repo?

chriskacerguis: @OBCENEIKON Yeah, we're ok here

stackus: Where did this cert come from? Wouldn't this mean every installation has the same certificate? How does an admin install a new certificate to lock down their installation?

chriskacerguis: It's a self signed cert.  It will be removed once we are done "testing" (long story as why we are "testing" with master).  Essentially, the cert is useless in any scenario other than this test.

PR #505, opened 2015-04-10
Bug/ux 419 project reject issues

tangerine007:
I'm not sure if this is the best way of refreshing view to show that project has been rejected. Also not sure about the :id/:project_id.


app/assets/javascripts/projects/project_controller.js
```diff
@@ -85,6 +85,7 @@ ProjectController.prototype = {
   },
   reject: function() {
     this.project.$reject({reason: this.reason});
+    location.reload();
```
stackus: This will cause the entire app to reload.

PR #511, opened 2015-04-13
Feature/remove settings

mafernando:
Removed references to old admin settings table.
Using Dotenv for these values.


chriskacerguis:
:+1:

PR #535, opened 2015-04-15
Feature/update sensu alerts

mafernando:
Removes attempts to add attributes to params var, instead populates @alert_params and @alert_id with before_actions and the proceeds accordingly.
To enable 'latest' to be a URL param Alerts could filter on, I needed to reduce the cyclomatic complexity of load_alerts - basically broke out each filter into its own private method.


chriskacerguis:
:+1:

PR #538, opened 2015-04-16
Feature/jel 329 modularize login system

OBCENEIKON:
This adds jellyfish-a-fied omniauth plugin compatibility.


app/models/staff.rb
```diff
@@ -63,4 +64,21 @@ class Staff < ActiveRecord::Base
   def gravatar
     '3fc88b95c85e43f157cb1ffd0e37e832'
```
chriskacerguis: I know that you didn't do this, but what is this for?

spec/factories/authentications.rb
```diff
@@ -0,0 +1,23 @@
...
+#  uid        :string
+#
+# Indexes
+#
+#  index_authentications_on_provider_and_uid  (provider,uid)
+#  index_authentications_on_staff_id          (staff_id)
+#
+
+FactoryGirl.define do
+  factory :authentications do
```
chriskacerguis: Do we need this FB stuff?

app/controllers/sessions_controller.rb
```diff
@@ -14,7 +14,15 @@ class SessionsController < Devise::SessionsController
   def create
     respond_to do |format|
       format.html do
-        super
+        if request.env['omniauth.auth']
+          auth_hash = request.env['omniauth.auth']
+
+          staff = Staff.find_by_auth(auth_hash)
+
+          sign_in_and_redirect(resource_name, staff) unless staff.nil?
```
calebthompson: double negative here... can you do `if staff.present?` instead?

app/models/staff.rb
OBCENEIKON: It looks like it's a method that just returns some static data. I found references to it in multiple places.
It looks like it's a method that just returns some static data. I found references to it in multiple places.
[config/routes.rb:34](https://github.com/projectjellyfish/api/blob/2429a1c45c05db1b729683812a158b590c13f969/config/routes.rb#L34)
[app/controllers/staff_controller.rb:17](https://github.com/projectjellyfish/api/blob/2429a1c45c05db1b729683812a158b590c13f969/app/controllers/staff_controller.rb#L17)
[doc/apipie_examples.json:3612, 3627, 3642](https://github.com/projectjellyfish/api/blob/2429a1c45c05db1b729683812a158b590c13f969/app/controllers/staff_controller.rb#L3612-3642)

I don't see a column for it in the staff table, you want me to remove it?

chriskacerguis: If we don't need it, then yeah let's clean it out (along with that other stuff)

OBCENEIKON:
Tested it, it does indeed work :)

PR #549, opened 2015-04-21
returns monthly projection as order total.

mafernando:
adds an `after_create` method to `order_items_controller` which will update parent `order` with the projected monthly cost of each service after the service is created.
projected monthly cost = setup + monthly + hourly * 750

### DIFF
--- a/app/models/order_item.rb
+++ b/app/models/order_item.rb
@@ -47,6 +47,9 @@ class OrderItem < ActiveRecord::Base
   has_many :provision_derivations
   belongs_to :latest_alert, class_name: 'Alert'
 
+  # Update the parent Order total
+  after_create :update_order_total
+
   # Hooks
   after_commit :provision, on: :create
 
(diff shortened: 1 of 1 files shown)


app/models/order_item.rb
```diff
@@ -47,6 +47,9 @@ class OrderItem < ActiveRecord::Base
   has_many :provision_derivations
   belongs_to :latest_alert, class_name: 'Alert'
 
+  # Update the parent Order total
+  after_create :update_order_total
```
mafernando: @stackus - this is a preliminary attempt at getting the total attribute on order populated. 

after talking with chris k and allen, we came to the decision that total on order should represent the monthly projected cost of that order. 

to accomplish this, i put an after_create action on order_items such that they update their parent order with their individual projected monthly cost each time one is created.

is there a better way to do this?

stackus: Yes, but for right now if this is working and it gets the job done it should be fine.

The other way would be a matter of refactoring things and using service objects or other tools to organize code in smaller classes or units of work. All the while chanting* "thin controllers and thin models".

*: chanting not actually required.

mafernando: :+1:  i like the sound of that chant, will keep that in mind going forward. thanks!

PR #558, opened 2015-04-22
Feature/fix project json

mafernando:
Makes `budget`, `spent`, `monthly_spend` return as JSON numbers for both zero and nonzero values.

### DIFF
--- a/db/migrate/20150422193059_change_spent_on_projects.rb
+++ b/db/migrate/20150422193059_change_spent_on_projects.rb
@@ -1,0 +1,6 @@
+class ChangeSpentOnProjects < ActiveRecord::Migration
+  def change
+    remove_column :projects, :spent
+    add_column :projects, :spent, :float
+  end
+end
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -51,7 +51,7 @@ class Order < ActiveRecord::Base
     grouped_order_items = order_items.group_by(&:project)
     grouped_order_items.reduce(false) do |over_budget, project_order_items|
       cost = project_order_items[1].map(&:calculate_price).sum
-      over_budget || cost + project_order_items[0].spent > project_order_items[0].budget
+      over_budget || cost + project_order_items[0].spent.to_f > project_order_items[0].budget
     end
   end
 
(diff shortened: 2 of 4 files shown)


mafernando:
I noticed that `budget` is defined as a float in the DB and was returning a JSON number regardless if the underlying value was zero or nonzero. 

Based on that observation, I got project `spent` to return as a JSON number by changing the project spent from a decimal to a float in the DB. 

Since `monthly_spent` comes from a method defined on the project model, I modified that method to cast the result as float which resulted in the response returning as JSON numbers for all cases.

chriskacerguis:
:+1:

db/migrate/20150422193059_change_spent_on_projects.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeSpentOnProjects < ActiveRecord::Migration
+  def change
+    remove_column :projects, :spent
+    add_column :projects, :spent, :float
```
stackus: See this article for why this is a bad idea.

http://spin.atomicobject.com/2014/08/14/currency-rounding-errors/

mafernando: @stackus - I'm looking into changing spent and budget back to decimals, but when I do they are returned as JSON strings in the request back to the client.

I've been looking through the concerns directory (`query_with_params` and `method/association_resolution`) as well as the serializers directory, but I'm having difficulty isolating where I can specify that those attributes should come back as JSON numbers instead of strings. Do you have any suggestions on where I should focus my attention?

stackus:
The answer here should fix it all up.

http://stackoverflow.com/questions/20006218/activesupport-encode-big-decimal-as-string

mafernando:
@stackus - [This](http://stackoverflow.com/questions/2593803/how-to-generate-the-create-table-sql-statement-for-an-existing-table-in-postgr) is brilliance. Couldn't figure out why decimals were coming back as JSON strings, now I know why!

Thanks for the course correct, I've changed spent and budget on project to decimals and then added the necessary config/application.rb, gemfile, and schema changes on PR https://github.com/projectjellyfish/api/pull/559

PR #559, opened 2015-04-23
Feature/fix project json

mafernando:
This fixes project json such that number attributes are not converted to strings when queried to the client.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -147,3 +147,6 @@ source 'http://rails-assets.org' do
   gem 'rails-assets-ngDraggable'
   gem 'rails-assets-angular-filter'
 end
+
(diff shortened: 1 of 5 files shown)


mafernando:
Implemented per Stack's feedback:

<table><tbody><tr><td>
See this article for why this is a bad idea.<br><br>

<a href='http://spin.atomicobject.com/2014/08/14/currency-rounding-errors'>http://spin.atomicobject.com/2014/08/14/currency-rounding-errors/</a><br><br>

If dealing with BigDecimal output is a hassle a better alternative is to use an integer and "move" the decimal point.
</tr></td></tbody></table>

chriskacerguis:
:+1:

PR #577, opened 2015-04-28
fix add services marketplace.

mafernando:
product type was previously looked up by id, but after refactor it should use keys from the product_types hash to link to products.


mafernando:
Next steps:
- [ ] update [product questions json](https://github.com/projectjellyfish/api/blob/master/config/product_questions/questions.json) with default product types

- [ ] map the existing products in the demo seed data to use the correct product type keys

app/assets/javascripts/projects/project_services_controller.js
```diff
@@ -19,7 +19,7 @@ var ProjectServicesController = function($scope, CartService, project, products,
 
   _.each(this.categories, _.bind(function(category) {
     category.products = _.filter(this.products, function(product) {
-      return product.product_type_id == category.id;
+      return product.product_type == category.title;
```
mafernando: ignoring this for now unless someone can suggest a different way to do this

calebthompson: We do this thrice. Here, in [list_projects_controller](https://github.com/projectjellyfish/api/blob/338490dcecad69ff51f221a9be2e9fe78e7f5afd/app/assets/javascripts/admin/products/list_products_controller.js#L10-14), and in [marketplace_controller](https://github.com/projectjellyfish/api/blob/338490dcecad69ff51f221a9be2e9fe78e7f5afd/app/assets/javascripts/marketplace/marketplace_controller.js#L10-14). Do you think we can extract this so that we can use a helper each time?

config/product_questions/application.json
```diff
@@ -0,0 +1,7 @@
+{
+  "Applications": {
```
calebthompson: I don't think we should keep these stub prodect types that match the demo seed products. Are they providing any value except visually for demos? They're going to pollute actual applications using this if they stick around, as they show up in the product_type dropdown on the form and can be used to create new products that can't be provisioned.

config/product_questions/database.json
```diff
@@ -0,0 +1,47 @@
+{
+  "Databases": {
+    "type": "object",
+    "title": "Databases",
```
calebthompson: We chatted about this a bit in person, but can we structure these so that the json-schema object is the root and map them to the title manually, like in [jellyfish_fog_aws](https://github.com/projectjellyfish/jellyfish_fog_aws/blob/master/config/initializers/product_types.rb#L4)?

config/product_questions/questions.json
```diff
@@ -47,78 +47,5 @@
         ]
       }
     }
-  },
-  "RDS": {
```
calebthompson: Why were the rest of these product types left?

app/assets/javascripts/projects/project_services_controller.js
mafernando: I think we should leave them as is in this PR because our main goal atm is to get JF demo ready for Thursday. I support the idea of refactoring this, but perhaps as part of a larger code clean-up in an upcoming sprint.

config/product_questions/application.json
mafernando: I believe leaving these stubbed out will be ok for now.

I've created Issue https://github.com/projectjellyfish/api/issues/581 to drive the resolution of determining what the default set of product types should be going forward.

@chriskacerguis - any thoughts here?

config/product_questions/questions.json
mafernando: I've tagged them for review/deletion on issue https://github.com/projectjellyfish/api/issues/581

config/product_questions/database.json
mafernando: I agree that we shouldn't be using two different approaches to structure the product types hash.

I've tagged this for resolution in issue https://github.com/projectjellyfish/api/issues/581.

chriskacerguis:
:+1:

Issue #590, opened 2015-04-30
project questions not appearing in project


mafernando:
why did these disappear?

![image](https://cloud.githubusercontent.com/assets/9356425/7417698/86006188-ef2d-11e4-865f-bf62f8413335.png)

![image](https://cloud.githubusercontent.com/assets/9356425/7417702/95620f3c-ef2d-11e4-87f5-3e26979e2ae4.png)

AllenBW:
don't entirely the question, the project details?

On Thu, Apr 30, 2015 at 12:39 PM, Michael Fernando <notifications@github.com

mafernando:
@AllenBW - shouldn't the answers to project questions appear when you view project details?

AllenBW:
Oh! i guess so? sure, that can be arranged.

On Thu, Apr 30, 2015 at 12:54 PM, Michael Fernando <notifications@github.com

chriskacerguis:
So, looks like the default /projects and /projects/:id no longer include that info.  You need to call

/api/v1/projects?includes[]=project_answers

(note the ?includes[]=project_answers)

mafernando:
looks like adding `project_answers` to [`'includes[]': ['approvals', 'approvers', 'services', 'staff']`](https://github.com/projectjellyfish/api/blob/master/app/assets/javascripts/projects/project_controller.js#L43) in the [ux project controller](https://github.com/projectjellyfish/api/blob/master/app/assets/javascripts/projects/project_controller.js) brings project questions back to the project details page. 

am trying to figure out why the values are blank.

![image](https://cloud.githubusercontent.com/assets/9356425/7421154/f8d7d4e2-ef47-11e4-8eb0-b6348cf4a789.png)

mafernando:
PR #597 fixes this - waiting on merge.

chriskacerguis:
Merged.

PR #597, opened 2015-04-30
return project questions by default.

mafernando:
fixes issue #590 and updates rake sample demo with some project questions


chriskacerguis:
Looks good.

PR #605, opened 2015-05-04
Bug/jel 569 type error undefined is not a function

AllenBW:


calebthompson:
So long as this fixes the problem in whatever browser had issues, :+1:

Issue #565, opened 2015-04-24
Add ability to add a Group/Groups when initiating a New Project


sonejah21:
Currently you can only add Groups when Editing a Project, on the project page itself. This should also be included in the Add Project form. 

I suggest using the same angucomplete-alt format, minus the heading, that is being developed in #427 and #512.

calebthompson:
Resolved by #606

Issue #512, opened 2015-04-13
Add Groups


calebthompson:
Group is an object that relates to Staff through Membership, and to Projects. Each of these are `has_many :through` relationships. Users will only have access to Projects through Groups (any existing direct relationships between User and Project will be replaced), which will vastly simplify permissions later.

Admins will create groups and add Staff and Projects to them.

chriskacerguis:
Dup of https://github.com/projectjellyfish/api/issues/485

stackus:
If you haven't considered it I'd suggest allowing the relationship to override the groups initial access level. This way when group access needs to be tweaked a duplicate but slightly different one need not be created. 

Also by still allowing adding a user to a project would allow for overriding any group membership access level they might have been given would further reduce creation of tweaked groups. This could be done the way it is now with some changes, or the membership relation can be altered to have one end use an abstraction or make it polymorphic.

A perfect example of a system which uses groups with no user access and no overrides of access is Github. Because it doesn't offer the flexibility to manage access separate from teams, and because it doesn't allow directly adding a user, owners create dozens of teams with a single user with various access levels that are then tied to different repos.

calebthompson:
Design in #427

sonejah21:
A question: Within the Project page, when you scroll down and currently see "Add User", and clicking it produces a modal with the search input field --- I assume that from our chat (and from the above in this ticket) a single User will no longer be able to be added to a Project. That said, you'd like me to change this "Add User" to "Add Group", and have the search input sift through the group names. The groups selected would then be added to the Project this way. Is that correct @calebthompson? Earlier this morning, @chriskacerguis and I spoke about it being "Add Users/ Groups". And when you'd search in the modal for a group or a user, both users and groups would be shown as possible results to add in. I just want to make sure that both of you are on the same page in terms of how we're handling users.

calebthompson:
Ah, I see the confusion. From the description above, which Chris said seemed right (but might have been missed):

> Users will only have access to Projects through Groups (any existing direct relationships between User and Project will be replaced), which will vastly simplify permissions later.

This makes the logic for connecting users and groups much simpler, and better symbolizes the idea that groups provide access to projects and roles/attributes provide abilities within those projects.

sonejah21:
Ahhh yes, ok. 

I believe you two were speaking the same language. Just wasn't sure when the omission of Add User on the Project page was taking place. For now, I'll simply add the /Group and include the ui-select inside. Hopefully that works for everyone :)

[...]

Issue #427, opened 2015-03-27
Create UX for group management


chriskacerguis:
Create concept of a group of users in the system.

- [x] Create groups in Admin
- [ ] Add / Remove users from groups
- [ ] Assign groups to projects

Core: https://github.com/projectjellyfish/api/issues/348

sonejah21:
Here's where it's at so far.... 

*Still need to flush out what information we need from an admin to initiate a group to build out the back end and finish the form. 

So far a group has the following data assigned to it via fake data streaming into list_groups_controller.js file: 
- id
- group_name
- group_email
- group_role (user group vs admin group - but we may not need this)
- created_at
- updated_at

When I can have a back-end person assist with this, it will also need to be able to contain the following:
- users
- user_emails
- user_roles
- others??? 

![uploads/735eb4a8-7766-46e7-89e1-e3f340f099f9/Screen Shot 2015-04-09 at 10.47.47 AM.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F735eb4a8-7766-46e7-89e1-e3f340f099f9%2FScreen+Shot+2015-04-09+at+10.47.47+AM.png)

chriskacerguis:
We'll need to add a column that will map the group to an LDAP group, or a SAML attribute.

sonejah21:
Alright cool, thanks! Sounds like back-end work, so I'll let the thoughtbot crew explain it to me/dig into this mapping once I am ready to hand off the front-end. @calebthompson @masonforest

sonejah21:
![uploads/1d3e9e9f-44b7-47d3-b4df-d00a5d8e33c9/add_groups_1.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F1d3e9e9f-44b7-47d3-b4df-d00a5d8e33c9%2Fadd_groups_1.png)

calebthompson:
Groups won't have roles at first, and admins should be able to add projects from group management, or groups from project management, whichever you prefer.

sonejah21:
@calebthompson Good idea - I'll add the ability to attach projects to a group into this page. Thanks for that! Since we'll eventually have the ability to add roles, I may just leave the Group Role in this page and disable it. Anything else this page needs?

calebthompson:
Looks good.

sonejah21:
REVISED "Add Groups" page: 
[...]

PR #606, opened 2015-05-04
Groups

calebthompson:


spec/support/features/selectize.rb
```diff
@@ -0,0 +1,6 @@
+module Features
```
calebthompson: This isn't used - should be removed.

spec/support/active_model_serializers_helper.rb
```diff
@@ -0,0 +1,24 @@
+module ActiveModelSerializersHelper
+  def match_serialized_json(object)
```
calebthompson: This is going to be super helpful making those request specs a little more robust. :+1: @masonforest

spec/requests/groups_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe 'Groups API' do
+  let(:default_params) { { format: :json } }
```
calebthompson: This is already present in half the specs. Maybe we should just remove it and move it into the actual specs?

spec/features/create_a_group_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    member_staff = create(:staff)
+    group = build(:group)
+    visit '/admin/groups/list'
+    click_on('Add New Group')
+
+    fill_in('Group Name', with: group.name)
+    fill_in('Group Description', with: group.description)
+    within('#users') do
+      find('.ui-select-search').click
+      find(:xpath, "//span[contains(text(),'#{member_staff.email}')]").click
```
calebthompson: This saddens me, but the selectize makes it the only option we could come up with. We could extract a method for it if anyone things that would clean things up...

spec/features/add_group_to_project_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    group = create(:group)
+    project = create(:project)
+    staff = create(:staff, :admin, groups: [group])
+    login_as staff
+
+    visit "/project/#{project.id}"
+    find('#group_search input').click
+    within('#group_search_dropdown') do
+      first('div', text: group.name).click
+    end
```
calebthompson: Again, sort of a weird edge case not captured by normal capybara helpers.

public/templates/partials/projects/project_form.html
```diff
@@ -49,6 +49,25 @@
     </div>
   </div>
 
+  <groups project="project"></groups> <!-- I think we want one of these instead -->
+  <section-header name="Groups">
```
calebthompson: @masonforest we got the groups-input working didn't we? I think we should plug it into the form as well.

public/templates/partials/projects/groups_input.html
```diff
@@ -0,0 +1,20 @@
+  <section-header name="Groups">
```
calebthompson: @sonejah21 this is what we came up with, along with https://github.com/projectjellyfish/api/pull/606/files#diff-7fbcbdb2590283819c54701ca411a1a0R1, for the groups search/listing on projects. It's going to be used on both project forms and on the project show page. What do you think?

calebthompson:
Seeing this on projects#show

app/assets/javascripts/groups/groups_resource.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+/**@ngInject*/
+var GroupsResource = function($resource, apiResource, $state) {
+    return $resource(apiResource('groupsById'), {'id': '@id'}, {
+        // Get Single
+        get: {
```
stackus: This is redundant. ngResource already sets up the get and query methods in the matter you're setting here.

app/controllers/projects_controller.rb
```diff
@@ -1,5 +1,5 @@
 class ProjectsController < ApplicationController
-  PROJECT_INCLUDES = %w(alerts approvals approvers latest_alerts project_answers project_detail services staff)
+  PROJECT_INCLUDES = %w(alerts approvals approvers latest_alerts memberships groups project_answers project_detail services)
```
calebthompson: Do we still need memberships? We may not since we're rendering groups directly for the most part.

app/assets/javascripts/admin/groups/admin_group_form_controller.js
```diff
@@ -0,0 +1,172 @@
+'use strict';
+
+var adminGroupsListState = 'base.authed.admin.groups.list';
+
+/**
+ * @todo This mirrors the product_form_controller.js file in a lot of regards, could be abstracted.
```
calebthompson: In a separate PR, I'd love to see a base controller we could extend that handled basic CRUD.

app/assets/javascripts/groups/groups_resource.js
calebthompson: Do we need this file at all then? I'm not sure what parts need to stick around.

app/assets/javascripts/projects/groups_input_directive.js
```diff
@@ -0,0 +1,40 @@
...
+        $scope.addGroupToProject = function(group) {
+          group = group.originalObject;
+          $scope.project.groups.push(group);
+          new MembershipsResource({
+            group_id: group.id
+          }).$save({projectId: $scope.project.id})
+        };
+
+        $scope.removeGroupFromProject = function(group) {
+          _.remove($scope.project.groups, {id: group.id});
```
stackus: Shouldn't this happen on success callback of the delete below? Or there needs to be an error handler that puts it back if it fails.

app/assets/javascripts/projects/project_controller.js
```diff
@@ -1,38 +1,25 @@
 'use strict';
 
 /**@ngInject*/
-function ProjectController($interval, project, ProjectUsersResource, OrderItemsResource, alerts, products, FlashesService) {
+function ProjectController($scope, $interval, project, OrderItemsResource, alerts, products, FlashesService, currentUser) {
 
-  this.intervalDelay = 30000;
-  this.$interval = $interval;
+  $scope.intervalDelay = 30000;
```
stackus: What's the reason for putting things onto $scope?

calebthompson: We found this on some angular sites, then found a counter-argument in the best practices guide you shared, but the idea is that we can do `{{ project.name }}` in the template rather than `{{ projectsCtrl.project.name }}` which seems much cleaner.

app/controllers/affiliations_controller.rb
```diff
@@ -0,0 +1,24 @@
+class AffiliationsController < ApplicationController
+  def create
```
stackus: Missing the api docs.

app/assets/javascripts/projects/groups_input_directive.js
calebthompson: Good point. My reasoning was that this would feel more reactive than waiting for the request. Which do you think makes more sense? Remove on click & re-add or remove in callback?

app/assets/javascripts/admin/groups/admin_add_group_controller.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
```
calebthompson: I'm fairly sure we can't do this because we're not wrapping the file in a closure. Should this warning be disabled? @stackus double check me?

app/assets/javascripts/projects/project_controller.js
calebthompson: WDYT?

app/assets/javascripts/groups/groups_resource.js
stackus: ngResource for facepalm reasons does not have an update method. So you'll want to keep that declaration.

app/assets/javascripts/admin/groups/admin_add_group_controller.js
stackus: This and most of the files can be made to use IIFEs with some changes

```
window.FooBar = (function(...) {
    'use strict';

    /** @ngInject */
    function FooBar(injected1, injected2) {};
    FooBar.resolve = {};
    return FooBar;
})(...);
```

Into `...` you can pass in `$`, `_`, or `moment` as needed. Most files should only be declaring one new thing so you can return that and have it attached to the window scope. If more than one thing needs to be declared then you _could_ pass in window as well and attach within the IIFE.

app/assets/javascripts/projects/groups_input_directive.js
stackus: Removing it immediately and then putting it back on error I think presents the better UX. Even if the task takes a long time, to the user it appears to have taken no time at all to complete.

app/assets/javascripts/admin/groups/admin_add_group_controller.js
calebthompson: Makes sense. That's a bit of a bigger change, and according to Chris K. we're going to replace the UX with CM's much nicer looking UX, so I think given our directive to do the minimum possible in the existing UX I'm going to ignore these.

app/assets/javascripts/projects/project_controller.js
stackus: The best practice is to put things onto the controller but the architecture of the JS app is so much a mixture of that practice and the old way of doing things that I would advocate not switching the styles either way. Though, I would suggest not switching to $scope if at all possible.

Current best practice is to use components so even the new way is the old way now too. Such is the world of JavaScript frameworks and architecture best practices. This week it'll be FRP or die as a dinosaur.

masonforest:
:+1:

calebthompson:
Resolves #512

calebthompson:
Resolves #565 and #427

PR #612, opened 2015-05-05
Authenticate users

masonforest:
We now authenticate users before authorizing them. This restricts controllers to signed in users only by default.


app/controllers/application_controller.rb
```diff
@@ -26,6 +26,11 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery
   skip_before_action :verify_authenticity_token, if: :json_request?, only: [:create, :acs]
+  before_action :authenticate_user
+
+  def authenticate_user
```
stackus: This is better named require_user. It does not actually authenticate the user.

masonforest: Good call -> ee8918ded4e19a9d73bd182513e90ccd9550d453

app/serializers/product_serializer.rb
```diff
@@ -10,11 +10,13 @@
...
 #  deleted_at           :datetime
+#  product_type_id      :integer
 #  setup_price          :decimal(10, 4)   default(0.0)
 #  hourly_price         :decimal(10, 4)   default(0.0)
 #  monthly_price        :decimal(10, 4)   default(0.0)
-#  provisioning_answers :jsonb
-#  product_type         :string
+#  provisionable_type   :string
+#  provisionable_id     :integer
+#  provisioning_answers :json
```
chriskacerguis: @masonforest What's the thought behind json vs jsonb?

masonforest: Oops `rake db:migrate` made these changes. Good catch. Reverted.

chriskacerguis:
:+1:

PR #627, opened 2015-05-11
Validate Role.permissions format

calebthompson:
Since this is a JSONb column and is pretty freeform, we want to ensure
the data we store in the permissions column follows the structure we
expect. Ensure that it contains only valid models and only valid
read/write actions. Also make sure we don't have garbage empty data
stored.


chriskacerguis:
:+1:

Issue #100, opened 2015-01-21
Add ability to have custom roles


chriskacerguis:
We need a way to add custom roles to the system.  So each page should have a "read" and "edit" (including actions) permission.

I'm thinking that this would work similar to the hook system.  We'd have a "is_allowed" on each method on controller, and then a permission system to match.  Need to discuss.

calebthompson:
Will edit and read for each class in the system be sufficiently granular? A not-real example might be publishing a blog post - should anyone able to author a post be able to publish one? A more realistic one might be approving a project - I know that's not something we want yet, so it'll have to be a little more granular I think than edit/read on objects.

read/edit on controllers make more sense, but since we don't have purely restful controllers that still may pose a concern since, for example, order_items have service start/stop and provision/retire actions that don't map to the same logic.

A thorough cleanup and extraction to only use restful controllers would help a lot with simplifying roles if that's a path we wanted to go down, but I don't think it would be a quick change.

calebthompson:
@masonforest any thoughts on any of this?

chriskacerguis:
I think edit / read is good.

calebthompson:
Okay so updated proposal based on a suggestion from @masonforest just now:

* read/write permissions on a per-controller level
* start with only project/product for now, add other controller-permissions later (this lets us get things merged faster and separate the work)
* create issues for other controllers that make sense to have as permissions
* Policies know to check permissions instead of blindly allowing/disallowing

@chriskacerguis is that good enough to get started on?

calebthompson:
I know the final boss here is project approvals, so I'd be down to throw that in as well.

![](http://i.ytimg.com/vi/M73ReyCWsnI/hqdefault.jpg)

calebthompson:
Strike that, products don't make sense in this context. The controllers that operate inside a project scope are:

projects
affiliations (of groups to projects)
approvals

So that's probably where it makes sense to start.

There are some controllers which should probably be operating in that context, so refactoring them to do so and incorporating them would be a good next step:

order_items (which currently include provisioning/retirement/start/stop, which should be refactored out and included in this list)
orders

calebthompson:
That leaves us with something like this for creating/editing roles:

![](http://caleb.click/atic/Image%202015-05-07%20at%2012.46.51%20PM.png)

calebthompson:
#619 and #624 builds roles and sets up permissions checking for projects, approve/reject, and memberships.

I think this is what's left:

[...]

PR #628, opened 2015-05-11
Add RolesController

calebthompson:
Supports basic actions for listing roles (to add them to groups) and for
admin CUD (no retrieval needed).

cc @sonejah21

#100


chriskacerguis:
@calebthompson looks like there is a rubocop issue, could you take a look?  Otherwise LGTM

calebthompson:
Fixed, merging.

PR #651, opened 2015-05-19
Add stack wizard frontend

calebthompson:
Users can now use a wizard to answer questions that will then dump them
into the add-service page with tags preselected (added to url bar,
they're not being used to filter since there's no filtering built into
the system).

* Create `<multiple-choice>` directive to facilitate switching between
  `input[type=radio]` for few possible answers and `select`s for many.
* Follow https://github.com/johnpapa/angular-styleguide for new files.


chriskacerguis:
Lgtm

PR #653, opened 2015-05-19
Add configuration for auto/manual wizard submit

calebthompson:
If $WIZARD_AUTOSUBMIT is false, wizard questions will be submitted with
an `input[type=submit]` rather than automatically on change.

### DIFF
--- a/app/assets/javascripts/common/constants.js
+++ b/app/assets/javascripts/common/constants.js
@@ -13,6 +13,8 @@ angular.module('broker.common.constants', [])
     orgColor: '#0a498a',
     version: '2.0.0'
   })
+  .constant('WIZARD_AUTOSUBMIT',
+      angular.element('meta[name="wizard-autosubmit"]').attr('content'))
   .constant('ROUTES', {
     login: '/login',
     logout: '/logout',
--- a/app/assets/javascripts/common/constants.js
+++ b/app/assets/javascripts/common/constants.js
@@ -1,23 +1,20 @@
 'use strict';
 
-window.appConfig = {
-  apiBasePath: window.apiBasePath,
-  orgLogo: '/images/logo.png',
-  orgColor: '#0a498a',
-  version: '2.0.0'
-};
 
-var CommonConstants = angular.module('broker.common.constants', [])
+angular.module('broker.common.constants', [])
   .constant('USER_ROLES', {
     all: '*',
     user: 'user',
     admin: 'admin'
   })
-  .constant('APP_CONFIG', angular.copy(window.appConfig))
+  .constant('APP_CONFIG', {
+    apiBasePath: angular.element('meta[name="api-base-path"]').attr('content'),
+    orgLogo: '/images/logo.png',
+    orgColor: '#0a498a',
+    version: '2.0.0'
+  })
   .constant('ROUTES', {
     login: '/login',
     logout: '/logout',
     default: '/login'
   });
-
-window.CommonConstants = CommonConstants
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/common/constants.js
```diff
@@ -1,23 +1,22 @@
 'use strict';
 
-window.appConfig = {
-  apiBasePath: window.apiBasePath,
-  orgLogo: '/images/logo.png',
-  orgColor: '#0a498a',
-  version: '2.0.0'
-};
```
calebthompson: I just inlined this since we weren't using `window.appConfig` anywhere (it was a relic of my poor understanding of the system during the ux merge).

app/views/welcome/index.html.erb
```diff
@@ -3,6 +3,8 @@
   <head>
     <!-- Meta Tags -->
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
+    <meta name="api-base-path" content="<%= root_url %>api/v1" />
+    <meta name="wizard-autosubmit" content="<%= ENV['WIZARD_AUTOSUBMIT'] %>" />
```
calebthompson: Instead of setting things on `window`, let Angular pull them from `meta` tags. Seems cleaner to me.

app/assets/javascripts/wizards/multiple_choice_directive.js
```diff
@@ -6,20 +6,29 @@
...
+
+  function MultipleChoiceDirectiveController($scope, WIZARD_AUTOSUBMIT) {
+    $scope.WIZARD_AUTOSUBMIT = WIZARD_AUTOSUBMIT;
+  }
+
+  function handleModelUpdate(scope, WIZARD_AUTOSUBMIT) {
+    if(WIZARD_AUTOSUBMIT) {
+      scope.$watch('model', function(newValue){
+        if(newValue) {
+          scope.action()
```
calebthompson: Basically two ways that `action` can be called. Either by this on('change') handler that happens if `WIZARD_AUTOSUBMIT == true` or in the `on('click')` handler in the template's `input[type=submit]` otherwise.

public/templates/partials/wizard/multiple_choice.html
```diff
@@ -16,4 +16,10 @@
     />
     {{option.text}}
   </label>
+  <input
+    type="submit"
+    value="Submit"
+    ng-click="action()"
```
masonforest: Can we rename `action` to `nextQuestion`?

public/templates/partials/wizard/multiple_choice.html
```diff
@@ -16,4 +16,10 @@
...
+    ng-if="!WIZARD_AUTOSUBMIT"
```
masonforest: How about`ng-hide="WIZARD_AUTOSUBMIT"` instead of a `!`?

public/templates/partials/wizard/multiple_choice.html
calebthompson: That's not meaningful in the context of a `<multiple-choice>`.

When it's passed into the directive, it is called that:

```html
<multiple-choice action="vm.nextQuestion"></multiple-choice>
```

public/templates/partials/wizard/multiple_choice.html
calebthompson: niiice.

app/assets/javascripts/wizards/multiple_choice_directive.js
AllenBW: thumbs, totally thumbs!

masonforest:
:+1:

PR #655, opened 2015-05-20
Wizard onepage

calebthompson:


public/templates/partials/wizard/multiple_choice.html
```diff
@@ -10,7 +10,7 @@
   >
     <input
       type="radio"
-      name="answer"
+      ng-name="option.question_id"
```
masonforest: can we make this `question-{option.question_id}`?

app/assets/javascripts/wizards/all_questions_controller.js
```diff
@@ -0,0 +1,31 @@
...
+      vm.answers = {};
+      vm.projectId = $stateParams.projectId;
+      vm.questions = questions;
+      vm.submitAnswers = submitAnswers;
+      vm.tags = [];
+      vm.$state = $state;
+
+      function submitAnswers() {
+        vm.questions.forEach(function(question) {
+          var answer = vm.answers[question.id];
```
masonforest: Where does `vm.answers` come in? Just curious.

public/templates/partials/wizard/multiple_choice.html
stackus: Out of curiosity are you using a module that provides `ng-name` or did you mean to use `ng-model`?

app/assets/javascripts/wizards/all_questions_controller.js
```diff
@@ -0,0 +1,31 @@
...
+    /**@ngInject*/
+    .controller('WizardAllQuestionsController', function($stateParams, $state, questions) {
+      var vm = this;
+
+      vm.answers = {};
+      vm.projectId = $stateParams.projectId;
+      vm.questions = questions;
+      vm.submitAnswers = submitAnswers;
+      vm.tags = [];
+      vm.$state = $state;
```
stackus: Looks like you can leave $state off the viewmodel. I mean unless you really enjoy typing `vm.` to access it, it's within scope below and can be accessed directly without issue.

public/templates/partials/wizard/multiple_choice.html
calebthompson: Neither. What's weird is that I was seeing behavior as if each radio had the same name and so all clicks reset other radios. That's done now, and I get a random name for each radio.

Fun...

app/assets/javascripts/wizards/all_questions_controller.js
calebthompson: It's populated [here](https://github.com/projectjellyfish/api/pull/655/files#diff-f5c696ad452296daaea8997d093bd6f0R7)

app/assets/javascripts/wizards/all_questions_controller.js
```diff
@@ -0,0 +1,30 @@
+(function() {
+  'use strict';
+
+  angular.module('broker.wizard')
+    /**@ngInject*/
+    .controller('WizardAllQuestionsController', function($stateParams, $state, questions) {
+      var vm = this;
+
+      vm.answers = {};
+      vm.projectId = $stateParams.projectId;
```
masonforest: Probably don't need to put this on the `vm` either.

masonforest:
:+1:

PR #656, opened 2015-05-21
Enhancement/ux 648 project form add groups input

sonejah21:
Needs edits to the spec file. Don't merge quite yet.


calebthompson:
General question, looks like you've re-indented all of the files to be four spaces instead of two. Intentional?

sonejah21:
Hmm. Unintentional. I only remember the indenting being off on a couple files, so I adjusted it. Not a necessary item I needed to check in.

public/templates/partials/projects/groups_input.html
```diff
@@ -1,20 +1,37 @@
...
-  <ul id="groups-list">
-    <li ng-attr-id="{{'group-' + group.id}}" class="group" ng-repeat="group in project.groups">
-      {{group.name}}
-      <select ng-model="role"></select>
-      <a href="#" ng-click="removeGroupFromProject(group)">x</a>
-    </li>
-  </ul>
-</section-header>
+    <div class="col-sm-12">
+        <groups-box class="product-box" ng-repeat="group in project.groups" group="group">
```
calebthompson: I don't think we want a group box here. I went with the `<ul>` because it gives us an attractive, minimal way to both remove and change roles for a membership (not for a group, as you've linked to).

I'm shooting for something a bit like this that allows us to set the role of a group, without needing to go to another page for just that one field.



Since groups don't have images, it also seems a little awkward to have the product box format for them since so little of the tile is meaningful (just the name and remove link).

spec/features/add_group_to_project_spec.rb
```diff
@@ -13,9 +13,10 @@
       first('div', text: group.name).click
     end
 
-    within('#groups-list') do
+    visit "/project/#{project.id}"
+    find('#group_search input').click
+    within('#group_search_dropdown') do
```
calebthompson: The purpose of this spec is to check that once we've added a group, it appears in the list of groups for the project. To do that, we want to check that it's either in a `.product-box` or that it is in the list as before (see other comment).

It's also useful to have the group id checked for so that we know we're not just added a different group with the same name (all groups have the same name in the factory). The name is almost optional.

spec/features/add_group_to_project_spec.rb
```diff
@@ -13,9 +13,10 @@
       first('div', text: group.name).click
     end
 
-    within('#groups-list') do
+    visit "/project/#{project.id}"
```
calebthompson: We shouldn't need to re-visit the page.

PR #660, opened 2015-05-26
update gemfile to vmware inf

mafernando:
updates lockfile to add ability to provision vmware clones with jellyfish-fog


chriskacerguis:
Lgtm

PR #662, opened 2015-05-26
Forward api requests through to the rails server

masonforest:
The goal is to have BrowserSync automatically reload the page whenever assets change.
 
Originally I tried proxying directly with BrowserSync -> http://www.browsersync.io/docs/options/#option-proxy

This forwarded ALL requests to the rails server. This won't work because the files rails serves in /public are static files. Once they have been generated with `gulp build` they won't be updated until we run `gulp build` again.

I think BrowserSync files will need to be served from Express.

This change proxys all api requests from the express app to the rails server.

You can run `gulp serve-dev` and run the rails server and see your assets update automatically.

I think eventually we'll want to get BrowserSync to work directly in rails. I think this will be fairly involved. Possibly [this gem](https://github.com/vigetlabs/gulp-rails-pipeline) will be helpful.


AllenBW:
:+1: :+1:

PR #667, opened 2015-05-28
Create Associations endpoint to add staff to group

### LINKED ISSUE #666: Improve Association API
Presently, associating a staff with a group is circuitous in that you need to get the full set of staff_ids, add or remove some number of ids from that, then PUT it back into the resource.

Set up an endpoint to POST create and DELETE destroy associations by staff id to simplify this.

calebthompson:
The previous approach, GETting a group, fiddling with its staff_ids
array, and PUTting it back, was awkward. This should prove a little more
straightforward.

Resolves #666, the neighbor of the pull request.


chriskacerguis:
:+1:

PR #668, opened 2015-05-28
Add Tags API

calebthompson:
* GET /tags for breakdown of all tags including counts.
* POST/DELETE /product/:product_id/tags to add or remove tags from a
  product.

@AllenBW You will probably need some of this soon.


chriskacerguis:
if passes. LGTM

PR #670, opened 2015-05-28
Enhancement/mf merge wizard into new ux

masonforest:
* I re-enabled some of the mocks so we can see the products marketplace with mocked data until we get the real data in there


masonforest:
Note: The test suite isn't passing in the new UX yet so this branch won't pass either.

chriskacerguis:
@masonforest can't auto-merge, can you take a look?

AllenBW:
:+1:

src/client/app/components/tags/tag-field.directive.js
```diff
@@ -259,7 +259,7 @@
 
       vm.newTag = '';
       vm.placeholder = '';
-      vm.tags = [];
+      vm.tags = vm.tags || [];
```
stackus: Why was this added? The field shouldn't have any tags applied to it directly. They should be coming from the parent directive.

src/client/app/core/constants.js
```diff
@@ -6,5 +6,7 @@
     .constant('lodash', _)
     .constant('jQuery', $)
     .constant('toastr', toastr)
+    .constant('WIZARD_AUTOSUBMIT', angular.element('meta[name="wizard-autosubmit"]').attr('content') == 'true')
```
stackus: Don't mix configuration constants in with application core constants.

src/client/app/components/api/api.constants.js
```diff
@@ -43,7 +43,7 @@
         'groupsById': '/groups/:id',
         'roles': '/roles',
         'rolesById': '/roles/:id',
-        'wizardQuestions': '/wizard_questions/:id'
+        'wizardQuestionsById': '/wizard_questions/:id'
```
stackus: Consider dropping this file completely. It only caused developers to look into two files to get the complete picture of resources endpoints. Each value is only ever used in a single resource anyhow.

src/client/app/mock-http-backend/config.js
```diff
@@ -5,6 +5,9 @@
     .run(mock);
 
   function mock($httpBackend) {
-    $httpBackend.whenGET(/^(?!\/api\/).+$/).passThrough();
+    $httpBackend.whenGET(/^(?!\/api\/mock\/).+$/).passThrough();
```
stackus: Feel like mocking things into special `/mock/` endpoints defeats the purpose of mocking things in the first place. Either the application intercepts the calls or it doesn't.

You're telling the application that all all `/api/mock` calls should go straight through to the real backend.

src/client/app/states/admin/wizard-questions/list/list.html
```diff
@@ -2,16 +2,91 @@
   <div class="admin-products-list__body">
```
stackus: This is using styles and classes from the admin-products styles.

The UI does not match with the rest of the new UI.

src/client/app/states/admin/wizard-questions/list/list.html
```diff
@@ -2,16 +2,91 @@
...
   </div>
-  <div class="admin-products-list__aside">
-    <div>
-      <product-type-show-modal
-        title="Add A Product"
-        message="Select the type of product you will be creating"
-        product-types="vm.productTypes"
-        on-ok="vm.createType(productType)"></product-type-show-modal>
+</content-header>
+<form class="form-horizontal padded horizontal-spanning">
```
stackus: Crutch classes `padded horizontal-spanning`.

src/client/app/states/admin/wizard-questions/list/list.html
```diff
@@ -2,16 +2,91 @@
...
+  <div ng-repeat="question in vm.questions">
+    <hr>
+    <div class="row">
```
stackus: Form controls should be put into `form-control` divs, and the labels and input use appropriate form-* classes. These should be BEM'ed as well.

src/client/app/states/marketplace/marketplace.state.js
```diff
@@ -14,7 +14,7 @@
   function getStates() {
     return {
       'marketplace': {
-        url: '/marketplace',
+        url: '/marketplace?tags',
```
stackus: Consider instead using ui-router params. See `app/states/admin/product/create/create.state.js` for an example.

src/client/app/states/marketplace/marketplace.state.js
```diff
@@ -45,14 +45,18 @@
...
-  function StateController(logger, $q, VIEW_MODES, CatalogService, Tag, Compare, TAG_QUERY_LIMIT) {
+  function StateController(logger, $q, VIEW_MODES, CatalogService, Tag, Compare, TAG_QUERY_LIMIT, $stateParams) {
     var vm = this;
 
     var COMPARE_LIMIT = 4;
 
     vm.title = 'Marketplace';
 
-    vm.tags = [];
+    if($stateParams.tags) {
```
stackus: This should be moved into activate().

src/client/app/states/marketplace/marketplace.state.js
```diff
@@ -45,14 +45,18 @@
...
+      vm.tags =  $stateParams.tags;
+    } else {
+      vm.tags = [];
```
stackus: More specifically this should be left here, and the existence check should be moved into activate.

src/client/app/states/wizard/multipage-wizard/multipage-wizard.state.js
```diff
@@ -0,0 +1,60 @@
...
+
+  /** @ngInject */
+  function appRun(routerHelper, navigationHelper) {
+    routerHelper.configureStates(getStates());
+  }
+
+  function getStates() {
+    return {
+      'wizard.multipage': {
+        controller: multiPageWizardController,
```
stackus: Use `StateController`.

src/client/app/states/admin/wizard-questions/list/list.state.js
```diff
@@ -16,9 +16,14 @@
       'admin.wizard-questions.list': {
         url: '', // No url, this state is the index of admin.products
         templateUrl: 'app/states/admin/wizard-questions/list/list.html',
-        controller: StateController,
+        controller: WizardQuestionsController,
```
stackus: Use `StateController`.

src/client/app/states/admin/wizard-questions/list/list.state.js
```diff
@@ -32,16 +37,56 @@
...
 
-    vm.title = 'Admin Products List';
-    vm.activate = activate;
+    vm.questions = questions;
+    vm.question = new WizardQuestion({wizard_answers: [{}]});
+    vm.createQuestion = createQuestion;
+    vm.addAnswer = addAnswer;
 
-    activate();
+    _.each(questions, addAnswer);
```
stackus: `lodash.each()`

lodash will need to also be injected into the controller. 

JSCS and JSHint should be telling you that this is a violation.

src/client/app/states/wizard/multipage-wizard/multipage-wizard.state.js
```diff
@@ -0,0 +1,60 @@
...
+    var vm = this;
+
+    vm.nextQuestion = nextQuestion;
+    vm.projectId = $stateParams.projectId;
+    vm.question = question;
+    vm.resetWizard = resetWizard;
+    vm.tags = [];
+
+    function nextQuestion() {
+      vm.tags = _.union(vm.tags, vm.answer.tags_to_add);
```
stackus: `lodash.union()`

src/client/app/states/wizard/multiple_choice_directive.js
```diff
@@ -0,0 +1,35 @@
+(function() {
```
stackus: This is in the wrong place.

src/client/app/states/wizard/multiple_choice_directive.js
```diff
@@ -0,0 +1,35 @@
...
+  'use strict';
+
+  angular.module('app.components')
+    .directive('multipleChoice', MultipleChoiceDirective)
+
+  function MultipleChoiceDirective() {
+    return {
+      controller: MultipleChoiceDirectiveController,
+      link: handleModelUpdate,
```
stackus: Use `link`.

src/client/app/states/wizard/multiple_choice_directive.js
```diff
@@ -0,0 +1,35 @@
...
+        action: "=",
+        model: "=",
+        options: "=",
+      },
+      templateUrl: 'app/states/wizard/multiple_choice.html',
+    }
+  };
+
+  function MultipleChoiceDirectiveController($scope, WIZARD_AUTOSUBMIT, WIZARD_MULTIPAGE) {
+    $scope.WIZARD_AUTOSUBMIT = WIZARD_AUTOSUBMIT;
```
stackus: There's no reason to put these values onto scope here. They're constants and are available always.

src/client/app/states/wizard/multiple_choice_directive.js
```diff
@@ -0,0 +1,35 @@
...
+      scope: {
+        action: "=",
+        model: "=",
+        options: "=",
+      },
+      templateUrl: 'app/states/wizard/multiple_choice.html',
+    }
+  };
+
+  function MultipleChoiceDirectiveController($scope, WIZARD_AUTOSUBMIT, WIZARD_MULTIPAGE) {
```
stackus: Don't use `$scope`.

src/client/app/states/wizard/multiple_choice_directive.js
```diff
@@ -0,0 +1,35 @@
...
+    $scope.WIZARD_AUTOSUBMIT = WIZARD_AUTOSUBMIT;
+    $scope.WIZARD_MULTIPAGE = WIZARD_MULTIPAGE;
+  }
+
+  function handleModelUpdate(scope) {
+    if(scope.WIZARD_AUTOSUBMIT && scope.WIZARD_MULTIPAGE) {
+      scope.$watch('model', function(newValue){
```
stackus: This `$watch` doesn't appear to be necessary. It looks like you're trying to react to the value changing instead of taking reacting to user action.

src/client/app/states/wizard/multiple_choice_directive.js
```diff
@@ -0,0 +1,35 @@
...
+  angular.module('app.components')
+    .directive('multipleChoice', MultipleChoiceDirective)
+
+  function MultipleChoiceDirective() {
+    return {
+      controller: MultipleChoiceDirectiveController,
+      link: handleModelUpdate,
+      restrict: 'E',
+      scope: {
+        action: "=",
```
stackus: Use `&?`. If you're going to be padding in function references use `&`. If the reference is optional add `?`.

src/client/app/states/wizard/singlepage-wizard/singlepage-wizard.state.js
```diff
@@ -0,0 +1,53 @@
...
+
+  /** @ngInject */
+  function appRun(routerHelper, navigationHelper) {
+    routerHelper.configureStates(getStates());
+  }
+
+  function getStates() {
+    return {
+      'wizard.singlepage': {
+        controller: singlePageWizardController,
```
stackus: `StateController`

src/client/app/states/wizard/wizard.state.js
```diff
@@ -0,0 +1,21 @@
...
+    .run(appRun);
+
+  /** @ngInject */
+  function appRun(routerHelper, navigationHelper, WIZARD_MULTIPAGE) {
+    routerHelper.configureStates(getStates(WIZARD_MULTIPAGE));
+  }
+
+  function getStates(WIZARD_MULTIPAGE) {
+    return {
+      'wizard': {
```
stackus: All of the child states of `wizard.` each resolve `questions`. That should instead be resolved here and not repeated for each child state.

src/client/app/states/wizard/wizard.state.js
```diff
@@ -0,0 +1,21 @@
...
+        url: "/project/:projectId/wizard",
+        redirectTo: WIZARD_MULTIPAGE ? 'wizard.multipage' : 'wizard.singlepage',
+        template: '<ui-view></ui-view>'
```
stackus: This can be dropped.

src/client/app/components/api/api.constants.js
masonforest: I agree! :+1:  I was following the convention here.

Let's discuss this change as a team and make a separate PR to remove this file if that's what we decide to do.

src/client/app/components/tags/tag-field.directive.js
stackus: I slept on this and think I now know why this was added. The actual init should happen in `activate()` though.

I know I've mentioned activate a lot and should explain why. The reason to put init, or startup code into activate is because eventually when the code is upgraded to Angular 2.0 and it uses ES6, the activates should be 1:1 transforms to the class constructor.

masonforest: Ah, gotchya. Thanks for the explanation: Fixed: [Move tags initialization to `activate()`](https://github.com/projectjellyfish/api/commit/3ac200183d11647d1340a3e7dc8a03b14a6bb19f)

src/client/app/mock-http-backend/config.js
masonforest: Yeah, I partially mocked the API to I could test how tags would work with the Wizard (the tags API isn't done yet). You're right though having a partially mocked API is confusing and having this unmocked will push us to finish the API: [Revert partial API mocking](https://github.com/projectjellyfish/api/commit/3d3f5927e436ce0fcbe9f76fd5c9453558aed0a3)

src/client/app/states/admin/wizard-questions/list/list.html
masonforest: The markup in the PR is just a place holder. I put minimal style on it to make it look not terrible. A designer will have to come through and make this pretty. At which point they can follow markup conventions, BEM, etc.

If we want I can try my hand a visual design but I wouldn't recommend it :flushed:

src/client/app/states/wizard/multipage-wizard/multipage-wizard.state.js
masonforest: Oh nice. I didn't notice we were doing this. [Fixed here and elsewhere](https://github.com/projectjellyfish/api/commit/31bc528aa5a9c73a36a4564148881706cb5e2996).

src/client/app/states/wizard/wizard.state.js
masonforest: Oh nice. I didn't know `resolves` worked across redirects. [Fixed](https://github.com/projectjellyfish/api/commit/3013fc30cc1b883ae556c4bbe44e685ad94cee5c)

src/client/app/states/wizard/wizard.state.js
masonforest: I wouldn't think I would need this either but when I remove it the child template fails to load.

`template` is also set to `<ui-view></ui-view>` [here](https://github.com/projectjellyfish/api/blob/enhancement/JEL_%23601-Integrate-CM-UX-into-Jellyfis-UX/src/client/app/states/projects/projects.state.js#L19).

masonforest:
https://www.dropbox.com/s/g6sa1kx6gy09cwh/jellyfish.mov?dl=0

src/client/app/states/wizard/multiple_choice_directive.js
masonforest: Hmmmm. Is there a way to access constants in the views without putting them on `$scope`?

I need access to them [here](https://github.com/projectjellyfish/api/pull/670/files#diff-67572dd8c26424f37f7f491e0841d740R34).

Relevant stackoverflow answer: http://stackoverflow.com/a/19383833

src/client/app/components/multiple-choice/multiple-choice.html
```diff
@@ -0,0 +1,39 @@
...
+        />
+      {{option.text}}
+    </label>
+    </div>
+  <input
+    class="btn btn-primary btn-lg pull-right"
+    type="submit"
+    value="Next"
+    ng-click="action()"
+    ng-hide="WIZARD_AUTOSUBMIT || !WIZARD_MULTIPAGE"
```
stackus: Instead of checking the constants you can do the check within the viewmodel and then do something like `vm.autoAdvance`

src/client/app/components/wizard/constants.js
```diff
@@ -0,0 +1,7 @@
+(function() {
+  'use strict';
+
+  angular.module('app.core')
```
stackus: module should be `app.components`.

src/client/app/components/multiple-choice/multiple-choice.html
masonforest: Cool. I set `autoSubmit` and `multiPage` on `vm` and then checked those in the view. -> https://github.com/projectjellyfish/api/commit/06abe633b6f05b3bf8a4ffe0b59e6352673f0778

PR #671, opened 2015-05-28
Record api examples to show in documentation

calebthompson:
cc @jklarfeld


calebthompson:
It’s less than ideal because we record _every_ api call in the tests, but until someone can do an audit/update of the :show_in_doc stuff that’s better than missing things.

chriskacerguis:
:+1: 

if passes

PR #672, opened 2015-05-29
Add input-wrap form directive

calebthompson:
This should DRY up a lot of the form building business.

### DIFF
--- a/public/templates/partials/common/question_input.html
+++ b/public/templates/partials/common/question_input.html
@@ -1,46 +1,22 @@
-<div ng-switch on="qiCtrl.question.field_type">
-
-  <div ng-switch-when="text" class="form-group" ng-class="{'has-error': qiCtrl.hasError()}">
-    <label class="col-sm-2 control-label" for="{{qiCtrl.fieldId()}}">{{qiCtrl.question.label}}</label>
-
-    <div class="col-sm-6 control-input">
+<input-wrap for="{{qiCtrl.fieldId()}}" label={{qiCtrl.question.label}} errorPresent="qiCtrl.hasError()">
+  <div ng-switch on="qiCtrl.question.field_type">
+    <div ng-switch-when="text" class="form-group" ng-class="{'has-error': qiCtrl.hasError()}">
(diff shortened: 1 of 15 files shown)


chriskacerguis:
lgtm

calebthompson:
I'm gonna wait on @sonejah21 and @AllenBW to see if they think this is a positive refactor.

public/templates/partials/common/question_input.html
```diff
@@ -1,46 +1,22 @@
-<div ng-switch on="qiCtrl.question.field_type">
-
-  <div ng-switch-when="text" class="form-group" ng-class="{'has-error': qiCtrl.hasError()}">
-    <label class="col-sm-2 control-label" for="{{qiCtrl.fieldId()}}">{{qiCtrl.question.label}}</label>
-
-    <div class="col-sm-6 control-input">
+<form-wrap for="{{qiCtrl.fieldId()}}" label={{qiCtrl.question.label}} errorPresent="qiCtrl.hasError()">
```
stackus: I think you meant input-wrap here.

public/templates/partials/forms/input_wrap.html
```diff
@@ -0,0 +1,4 @@
+<div class="form-group" ng-class="{'has-error': errorPresent}">
+  <label class="col-sm-3 control-label" for="{{for}}">{{label}}</label>
```
stackus: These two values should be once-bound. To do this add `::` to the start of the variable

example: `{{::for}}`

calebthompson: Just in this template, or everywhere the directive is used?

public/templates/partials/common/question_input.html
calebthompson: Sure did, thanks!

public/templates/partials/forms/input_wrap.html
stackus: Just here. Doing this should reduce the number of additional watches by two every time this directive is used.

calebthompson:
@stackus thanks. I addressed both of those points.

masonforest:
:+1:

PR #679, opened 2015-06-01
Introduce and use broker-input where appropriate

calebthompson:
`<broker-input>` is a shortcut for an `<input>` wrapped in an
`<input-wrap>` directive and handles the simplest (and most common) case
for its use.

A likely future improvement here would be either to introduce
`<broker-textarea>` and `<broker-select>` for those form controls, but I
think this change itself is valuable.


calebthompson:
@sonejah21 @stackus @masonforest

masonforest:
lgtm

sonejah21:
This is very cool - thanks Caleb & Mason. I'll start the styling changes for the list of groups after it gets merged in.

PR #680, opened 2015-06-02
Enhancement/ux 648 new groups input styling

sonejah21:
This adds basic styling to the list of groups selected. 
@calebthompson any idea why the groups input field vanishes once a group is selected?


calebthompson:
Screenshot? I don't see that happening on master:

Not happening

sonejah21:
Here are some screenshots... 
screen shot 2015-06-02 at 11 02 51 am

screen shot 2015-06-02 at 11 01 57 am

app/assets/stylesheets/projects/_styles.sass
```diff
@@ -186,3 +186,19 @@ tbody
       .text-undo
         font-weight: 400
         color: blue
+
+.list-group-block
+  a
+    .fr
+      float: right
+  .list-group-item
+    padding: 20px 10px 70px
```
calebthompson: This is a lot of top/bottom padding IMO.

calebthompson:
@sonejah21 interesting. Which page is that on? Is it happening for you on that page on master?

sonejah21:
On master... Create a group or two, and then go to the /project/:id page. Add a group you created to a project. On my version of master, it flashes on top of the input. I tried adjusting the padding but it shows nothing on top, just white space, as if the input were not there anymore.

calebthompson:
Might be a local thing. I attempted to reproduce on https://jellyfish-core-dev.herokuapp.com/project/4 and couldn't.

I did notice that the form styles aren't working on that page. Possibly we'll need to wrap the element in a `<form>` tag or adjust the css to be less specific.

sonejah21:
You're right about it working on the heroku version. I'll look into this more on my end. 

Can you attach a screenshot of which form isn't being styled correctly? The only styling thing I see needing attention is the groups list that I just did.

calebthompson:
On the page I linked, the input-wrap stuff doesn't seem to be inheriting the same styles as elsewhere, since it's not inside a form tag.

calebthompson:
No idea if that's been addressed in this branch but it seems not to have been since I don't see any changes to that markup.

PR #682, opened 2015-06-02
Enhancement/681 documentation routing

sonejah21:
Added in:
- System Requirements file
- INSTALL-OSX file
- INSTALL-RHEL file

These are needed to link properly from the landing page/ README file.


calebthompson:
Should these be removed from the root of the project instead of copied?

sonejah21:
Since they currently live on the Wiki page in Github - which hasn't been moved over yet to the ReadtheDocs (via the /docs folder) - I left them in the Wiki for the time being. Once the Wiki pages are moved into a nested format in RTD then I was going to remove them all from the Wiki. 

A dumb question though: does the Wiki of a repo live hidden in the root of a project? I thought those files lived on the Github GUI alone.

calebthompson:
The two install files aren't: https://github.com/projectjellyfish/api/blob/master/INSTALL-OSX.md and https://github.com/projectjellyfish/api/blob/master/INSTALL-RHEL.md

sonejah21:
Fixed :+1: The two install files were removed from root.

calebthompson:
:+1:

PR #685, opened 2015-06-03
Temporarily fix tag_list in API

calebthompson:
This should be reverted and the underlying problem solved, which is that
the `taggable_tags` stuff is not creating regular, unscoped tags
anymore.

If we're just trying to get tags by product_type, I think there's
probably SQL we can do to make that happen a little more cleanly. Ping
me.

### DIFF
--- a/app/serializers/product_serializer.rb
+++ b/app/serializers/product_serializer.rb
@@ -22,9 +22,13 @@
 #
 
 class ProductSerializer < ApplicationSerializer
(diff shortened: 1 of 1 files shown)


calebthompson:
@OBCENEIKON @AllenBW

chriskacerguis:
LGTM if pass

calebthompson:
@AllenBW can you check out this branch and see if it's working for you (tag_list should be populated again in the API.

stackus:
Needs comments in the code to explain the 'temporary' nature of the changes.

PR #699, opened 2015-06-10
Feature/add tags url param to products

mafernando:
adds tags url parameter to products


AllenBW:
:+1: YES LETS BRING IT ON!!!!!!

PR #701, opened 2015-06-11
Bug/jel 641 required product attributes

### LINKED ISSUE #641: Products don't require attributes for creation
When creating a new product from a product type, there are required questions that need to be answered. Current implementation of API does not seem to honor these requirements and will allow you to create a product with missing attributes.

#### Creating a New Product
image

#### From a Product Type with pre-defined questions
image

### Does not honor the 'required' attribute and allows you to create products if required attributes are not present
image

OBCENEIKON:
Fixes #641

### DIFF
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -35,11 +35,15 @@ class ProductsController < ApplicationController
   def create
     product = Product.new(product_params)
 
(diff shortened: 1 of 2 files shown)


mafernando:
@OBCENEIKON - the way this was originally set up was requiring that all provisioning answers be required, even the optional ones. The latest commit pulls in the list of required attributes for the product type from the global namespace and only checks if those attributes are present for new product creation.

chriskacerguis:
:+1:

PR #707, opened 2015-06-12
Enhancement/681 documentation nesting

sonejah21:


docs/Wiki/Modules/Steps_1_to_7: Jellyfish_Module_Setup.md
```diff
@@ -1,3 +1,5 @@
+# Jellyfish Module Setup (Steps 1-7)
```
mafernando: Hmm. There are 1-9 steps in the module guide and step 8 is optional. Maybe this could be confusing?

sonejah21: Maybe we could do basic steps as the Guide and then add the optional step to Add, describing where it would come into play before the final step.

mafernando: Works for me!

PR #713, opened 2015-06-16
Reenable notional charts

AllenBW:


stackus:
Instead of dumping the data into the controller, use a resource/model or hold it and inject it in.

PR #714, opened 2015-06-16
Fixes/orders placed by

stackus:


stackus:
Don't merge yet.

AllenBW:
niiiiiiiiiiiiiice updated controller means we don't need a service to grab usernames!

would also like to add project name to alerts if possible... for giggles n such

PR #728, opened 2015-06-16
Bug/remove public

mafernando:
remove public dir


chriskacerguis:
Something not right with this

Issue #749, opened 2015-06-16
Reordering project questions doesn't save


OBCENEIKON:
Attempting to reorder project questions doesn't save, though it says that it has.

PR #759, opened 2015-06-17
Fixed reordering issue on the API side for project questions. Also re…

OBCENEIKON:
…factored the code to use the latest style.

This fixes the API side of things for #749, there's still a UX bug I was trying to fix. Anytime a question gets reordered, it's only updating load_order for the selected question so two questions will end up with the same load_order.
Not sure if this should be a fix on the UX side or on the API side...


chriskacerguis:
Can you work on the UX side and update the pull request when you are done?

OBCENEIKON:
Will do.

chriskacerguis:
Going to close.  Please redo PR when you have the UX done too

PR #762, opened 2015-06-19
Bug/743 roles ui

AllenBW:


app/controllers/roles_controller.rb
```diff
@@ -1,42 +1,59 @@
...
-    authorize(roles)
     respond_with_params roles
   end
 
+  api :GET, '/roles/:id', 'Returns roles with :id'
+  param :id, :number, required: true
+  error code: 404, desc: MissingRecordDetection::Messages.not_found
+
+  def show
+    respond_with_params role
```
stackus: `respond_with` is better here since the request doesn't allow for any parameters with uses the `_params` part.

app/controllers/roles_controller.rb
```diff
@@ -1,42 +1,59 @@
...
+  end
+
   api :POST, '/roles', 'Create a role'
   document_params required: true
+
   def create
-    authorize(Role)
-    respond_with_params Role.create!(role_params)
+    authorize Role
+    respond_with Role.create role_params
```
stackus: Create the role, then authorize and return.

app/models/role.rb
```diff
@@ -14,6 +14,10 @@ class Role < ActiveRecord::Base
   has_many :memberships
   has_many :projects, through: :memberships
 
+  serialize :permissions, HashSerializer
```
stackus: `serialize` is deprecated.

app/serializers/hash_serializer.rb
```diff
@@ -0,0 +1,9 @@
+class HashSerializer
```
stackus: This shouldn't be needed.

app/models/role.rb
```diff
@@ -29,7 +33,7 @@ def all_valid_keys
   end
 
   def all_valid_values
-    unless permissions.values.all? { |perms| perms & %w(read write) == perms }
+    unless permissions.values.all? { |perms| perms & %w(read write ) == perms }
```
stackus: extra space

src/client/app/components/role-form/role-form.directive.js
```diff
@@ -0,0 +1,102 @@
...
+
+      vm.backToList = backToList;
+      vm.showErrors = showErrors;
+      vm.hasErrors = hasErrors;
+      vm.onSubmit = onSubmit;
+
+      function activate() {
+        // required to build permissions list
+        vm.permissions = {approvals: {}, projects: {}, memberships: {}};
+        lodash.each(vm.role.permissions, function(value, key) {
```
stackus: No inline/anonymous functions.

src/client/app/components/role-form/role-form.directive.js
```diff
@@ -0,0 +1,102 @@
...
+      vm.showValidationMessages = false;
+
+      vm.backToList = backToList;
+      vm.showErrors = showErrors;
+      vm.hasErrors = hasErrors;
+      vm.onSubmit = onSubmit;
+
+      function activate() {
+        // required to build permissions list
+        vm.permissions = {approvals: {}, projects: {}, memberships: {}};
```
stackus: The permissions should be part of the role model.

src/client/app/components/role-form/role-form.directive.js
```diff
@@ -0,0 +1,102 @@
...
+
+        return vm.showValidationMessages && vm.form[field].$invalid;
+      }
+
+      function onSubmit() {
+        vm.showValidationMessages = true;
+
+        // to parse permissions in proper form, undefined throws validation errors
+        lodash.each(vm.permissions, function(value, key) {
+          vm.temp = [];
```
stackus: A comment.

Not every value needs to be on the viewmodel. Something like a temp value is better scoped to the function that uses it.

src/client/app/components/role-form/role-form.html
```diff
@@ -0,0 +1,140 @@
...
+                    placeholder="Enter a role description." required></textarea>
+
+          <div class="role-field__feedback" ng-messages="vm.form.description.$error">
+            <div ng-message="required">An role description is required.</div>
+          </div>
+        </div>
+      </div>
+
+
+      <div class="role-field">
```
stackus: This list can be reduced with an ng-repeat.

src/client/app/resources/roles.factory.js
```diff
@@ -6,8 +6,25 @@
...
+        method: 'PUT',
+        isArray: false
+      }
+    });
 
-    return Roles;
+    Role.defaults = {
+      name: '',
+      description: '',
+      permissions: {approvals: {}, projects: {}, memberships: {}}
```
stackus: Verify that the permissions object is copied. You might find that it's holding onto changes inbetween edits and creations..

src/client/app/states/admin/roles/create/create.html
```diff
@@ -1 +1 @@
-<product-form product="vm.product" product-type="vm.productType"></product-form>
+<role-form role="vm.role" heading="Create Role"></role-form>
```
stackus: 'Create A Role'

src/client/app/states/admin/roles/list/list.html
```diff
@@ -28,16 +28,18 @@
     </tr>
     </thead>
     <tbody>
-    <tr class="roles-table__row" ng-repeat="row in vm.roles">
+    <tr class="roles-table__row" ng-repeat="row in vm.role">
       <td class="roles-table__cell roles-table__cell--padding"></td>
       <td class="roles-table__cell roles-table__cell--order">{{ ::row.id }}</td>
       <td class="roles-table__cell roles-table__cell--question">{{ ::row.name}}</td>
```
stackus: It's very important to class the columns correctly. Giving all but one column a fixed width will keep the table from resizing columns when columns are sorted.

src/client/app/states/admin/roles/list/list.html
```diff
@@ -7,13 +7,13 @@
...
-  <h4>No roles presently exist.  Add a new one!</h4>
+<div ng-if="!vm.role.length">
+  <h4>No roles presently exist. Add a new one!</h4>
 </div>
-<div class="roles-table" ng-if="vm.roles.length">
+<div class="roles-table" ng-if="vm.role.length">
   <table st-table="vm.roles" class="roles-table__table table-hover">
     <thead>
-    <th colspan="3">
+    <th colspan="4">
```
stackus: Nix this filter box.

src/client/app/states/admin/roles/list/list.html
```diff
@@ -28,16 +28,18 @@
...
     <tbody>
-    <tr class="roles-table__row" ng-repeat="row in vm.roles">
+    <tr class="roles-table__row" ng-repeat="row in vm.role">
       <td class="roles-table__cell roles-table__cell--padding"></td>
       <td class="roles-table__cell roles-table__cell--order">{{ ::row.id }}</td>
       <td class="roles-table__cell roles-table__cell--question">{{ ::row.name}}</td>
       <td class="roles-table__cell roles-table__cell--question">{{ ::row.description }}</td>
       <td class="roles-table__cell roles-table__cell--question">
-        <span ng-repeat="permission in row.permissions">{{permission}}, </span>
+        <span ng-repeat="(category, permissions) in row.permissions" ng-if="permissions"><b>{{ ::category}}</b>:
```
stackus: These nested ng-repeats are unnecessary. A resource or perhaps a viewmodel function which returns formatted text would be a better option.

src/client/app/states/admin/roles/list/list.state.js
```diff
@@ -20,7 +20,7 @@
         controllerAs: 'vm',
         title: 'Admin Roles List',
         resolve: {
-          roles: resolveRoles
+          role: resolveRoles
```
stackus: The plural form made more sense.

app/controllers/roles_controller.rb
```diff
@@ -1,42 +1,59 @@
 class RolesController < ApplicationController
   def self.document_params(required: false)
-    param :name, String, desc: 'Name of the role', required: required
+    param :name, String, desc: 'Name of the role'
```
stackus: Name should be required.

src/client/app/states/admin/roles/list/list.state.js
AllenBW: addressed

src/client/app/states/admin/roles/list/list.html
AllenBW: Got no problem doing this, but in the cases where there are many roles
listed would this not be a valuable user convenience method?  What's the
cause for nixing?

stackus: No other reason other than we're not including filter boxes with tables.

AllenBW: addressed

src/client/app/states/admin/roles/list/list.html
AllenBW: addressed

src/client/app/states/admin/roles/create/create.html
AllenBW: addressed

src/client/app/resources/roles.factory.js
AllenBW: addressed

src/client/app/components/role-form/role-form.directive.js
AllenBW: addressed

src/client/app/components/role-form/role-form.directive.js
AllenBW: addressed

src/client/app/components/role-form/role-form.directive.js
AllenBW: addressed

src/client/app/components/role-form/role-form.html
AllenBW: addressed

app/models/role.rb
AllenBW: addresssed

app/controllers/roles_controller.rb
AllenBW: addressed

src/client/app/states/admin/roles/list/list.html
AllenBW: addressed

src/client/app/states/admin/roles/list/list.state.js
```diff
@@ -73,5 +67,14 @@
         Toasts.error('Server returned an error while deleting.');
       }
     }
+
+    function permissionList(input) {
```
stackus: Consider using `.map()`

```
function permissionsList(list) {
  return lodash.flatten(lodash.map(list, formatPermissions)).join('');

  function formatPermissions(value, key) {
    return ['<strong>', key, '</strong>:', value.join(' '), <br>'];
  }
}
```

That should work and is also an example of what I was referring to when I said to pull out the inline/anonymous function.

src/client/app/states/admin/roles/list/list.html
```diff
@@ -8,36 +8,28 @@
...
-             class="form-control roles-table__filter-input"
-             placeholder="Filter by name, description, permissions"
-             type="text"/>
-    </th>
     <tr>
       <th class="roles-table__heading roles-table__heading--padding"></th>
       <th class="roles-table__heading roles-table__heading" st-sort="id">Id</th>
-      <th class="roles-table__heading roles-table__heading--question" st-sort="name">Name</th>
-      <th class="roles-table__heading roles-table__heading--required" st-sort="description">Description</th>
+      <th class="roles-table__heading roles-table__heading" st-sort="name">Name</th>
```
stackus: If the column is sortable the `--sortable` modifier should be added.

src/client/app/states/admin/roles/list/list.state.js
AllenBW: thanks for this! :100: :+1:

src/client/app/states/admin/roles/list/list.html
```diff
@@ -8,36 +8,28 @@
...
     <tbody>
     <tr class="roles-table__row" ng-repeat="row in vm.roles">
       <td class="roles-table__cell roles-table__cell--padding"></td>
-      <td class="roles-table__cell roles-table__cell--order">{{ ::row.id }}</td>
-      <td class="roles-table__cell roles-table__cell--question">{{ ::row.name}}</td>
-      <td class="roles-table__cell roles-table__cell--question">{{ ::row.description }}</td>
-      <td class="roles-table__cell roles-table__cell--question">
-        <span ng-repeat="permission in row.permissions">{{permission}}, </span>
-      </td>
+      <td class="roles-table__cell roles-table__cell">{{ ::row.id }}</td>
```
stackus: Assign each column a `--{column-iden}` and make the id, name, and actions have a fixed width so we don't see any wiggle when the columns are sorted.

PR #773, opened 2015-06-26
Feature/travisci cache dependencies

stackus:
@chriskacerguis You've messed with this the most, so if you'd care to have a look I'd appreciate it.


chriskacerguis:
@stackus looks good, but let's see if the build passes.

stackus:
Still making small edits. After testing if the vendor/bundle caches I'm going to remove a redundant `gulp build` which will also shave off 40s to 1m worth of build time.

PR #778, opened 2015-06-30
Add token authentication strategy

### LINKED ISSUE #763: Token Authentication is non-functional/changed.
It took me a while to notice this, but despite the fact that/staff/sign_in does indeed return a token, it cannot be used.

When using this token in the `X-STAFF-TOKEN` header field in a request (code unchanged from before) every request returns a 401.  

Apparently, the service still returns a cookie and will favor that over the token, even if a token is present.  Once you completely disable cookie acceptance in the iOS code base, this issue becomes noticeable, as apparently Cookie auth was working transparently for a while.  When you redeploy an iOS app however, all cookies are removed, preventing you from being able to authenticate again.

So either token auth is completely non-functional,

**OR**

The header key has changed, and is no longer `X-STAFF-TOKEN`

stackus:
- Remove the TokenAuthentication concerns for the controller and model
- Use non-bang versions of create so that the API returns more json like error responses
- Create token on login
- Delete token on logout
- Remove authentication_token from staff modal
- Minor edits to the JS specs to support integration testing

Fixes #763


app/models/alert.rb
```diff
@@ -2,22 +2,22 @@
 #
```
chriskacerguis: What's the context here?  (why changing the alerts stuff)

stackus: I guess master didn't have the updated annotations from the alerts refactor.

PR #783, opened 2015-06-30
Feature/update alerts ux

mafernando:
Refactors admin alerts to be where organization (system-wide) alerts get posted/edited/viewed.
Alert create/edit for projects/staff/order_items will be handled from respective details pages.
Some issues with editing a project still persist, am investigating.

### DIFF
--- a/app/controllers/alerts_controller.rb
+++ b/app/controllers/alerts_controller.rb
@@ -108,6 +108,6 @@ class AlertsController < ApplicationController
 
   def alert_params
     @_alert_params ||= params.permit(:id, :status, :message, :category, :start_date, :end_date)
-    @_alert_params = @_alert_params.merge({ alertable_type: Organization, alertable_id: 1}) if request.path == '/api/v1/alerts'
+    @_alert_params = @_alert_params.merge(alertable_type: Organization, alertable_id: 1) if request.path == '/api/v1/alerts'
   end
 end
--- a/app/controllers/alerts_controller.rb
+++ b/app/controllers/alerts_controller.rb
@@ -99,6 +99,6 @@ class AlertsController < ApplicationController
   end
 
   def alert_params
-    @_alert_params ||= params.permit(:status, :message, :category, :start_date, :end_date, :alertable_type, :alertable_id)
+    @_alert_params ||= params.permit(:id, :status, :message, :category, :start_date, :end_date, :alertable_type, :alertable_id)
   end
 end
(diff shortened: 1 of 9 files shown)


chriskacerguis:
Looks like there are some issues with the build.  @mafernando can you please address

app/controllers/alerts_controller.rb
```diff
@@ -91,6 +99,6 @@ def apply_not_status(query)
   end
 
   def alert_params
-    @_alert_params ||= params.permit(:status, :message, :category, :start_date, :end_date)
+    @_alert_params ||= params.permit(:id, :status, :message, :category, :start_date, :end_date, :alertable_type, :alertable_id)
```
stackus: We shouldn't accept the alertable_* columns. That would allow alerts to be maliciously reassigned.

app/serializers/alert_serializer.rb
```diff
@@ -21,6 +21,6 @@
 #
 
 class AlertSerializer < ApplicationSerializer
-  attributes :id, :status, :message, :category
+  attributes :id, :status, :message, :category, :alertable_type, :alertable_id
```
stackus: Likewise, we shouldn't expose the alertable_* columns.

src/client/app/components/alert-form/alert-form.directive.js
```diff
@@ -55,11 +55,12 @@
       vm.openAnswerDate = openAnswerDate;
 
       function activate() {
+        // ONLY SUPPORT POSTING TO SINGLE ORG FOR NOW
+        vm.alertToEdit.alertable_type = 'Organization';
```
stackus: Have this defaulting happen on the server.

src/client/app/states/admin/alerts/list/list.html
```diff
@@ -25,10 +24,9 @@
         </tr>
         </thead>
         <tbody>
-        <tr class="alerts-table__row" ng-repeat="row in vm.alerts">
+        <tr class="alerts-table__row" ng-repeat="row in vm.alerts | filter: alertable_type = 'Organization'">
```
stackus: Here, you can avoid having to filter the list by having the alerts index route only send the Org alerts.

app/controllers/alerts_controller.rb
mafernando: addressed - removed alertable* from the controller

app/serializers/alert_serializer.rb
mafernando: addressed - removed alertable* from the serializer

src/client/app/components/alert-form/alert-form.directive.js
mafernando: addressed - moved this logic into params for the alerts controller

src/client/app/states/admin/alerts/list/list.html
mafernando: addressed - using url_param to specify the result from api comes back with only organization alerts.

PR #784, opened 2015-07-01
Added group members to groups object.

OBCENEIKON:


app/models/group.rb
```diff
@@ -13,4 +13,5 @@ class Group < ActiveRecord::Base
   has_and_belongs_to_many :staff
   has_many :memberships
   has_many :projects, through: :memberships
+  has_many :groups_staffs
```
stackus: Staff is already a relationship. It is defined there on line 13.

app/models/project.rb
```diff
@@ -38,6 +38,7 @@ class Project < ActiveRecord::Base
   has_many :memberships
   has_many :groups, through: :memberships
   has_many :staff, through: :groups
+  has_many :groups_staffs, through: :groups, source: :staff
```
stackus: GroupStaff is redundant here as well. The staff relationship for projects is declared up one line on 40.

app/serializers/alert_serializer.rb
```diff
@@ -21,6 +21,7 @@
 #
 
 class AlertSerializer < ApplicationSerializer
-  attributes :id, :status, :message, :category
+  attributes :id, :status, :message
   attributes :start_date, :end_date, :created_at, :updated_at
+  attributes :alertable_id, :alertable_type, :category
```
stackus: Shouldn't be sending the alertable_* columns.

OBCENEIKON: Those were added because the rspec tests were failing. Michael later informed me they needed to be removed, so I fixed the rspec test instead :)

Issue #785, opened 2015-07-01
Editing Alerts doesn't work


mafernando:
When editing an alert, the form sends the values of the alert as it is queried from the DB, not how it exists after being edited in the alert form. Consequently, the alert that is saved never changes attributes.

mafernando:
Fixed via PR https://github.com/projectjellyfish/api/pull/786.

PR #786, opened 2015-07-01
Feature/fix edit admin alerts

mafernando:
address Issue https://github.com/projectjellyfish/api/issues/785
adds edit state (thanks @AllenBW for the inspiration from projects) to use for editing exiting organization alerts.


AllenBW:
:+1:!!!!

Issue #194, opened 2015-02-05
Restrict alerts to display only those where status is not OK


mafernando:
Restrict alerts to display only those where status is not OK

stackus:
The reason is you're making API work for a single client by doing that. Some client may expect that the `all_alerts` call will do just that, return all alerts. An open API such as this shouldn't assume what a client is going to want.

What should be done is the following.

- Make GET /alerts return an unfiltered list of alerts
- Add parameters to let a client choose to modify that list
    - Options such as active=true, status=[OK, CRITICAL,...], sort=[project, oldest]
- Create the logic to add the scopes requested by the client
- Remove the extra GET /alert_* actions
- Update UX to request the data it wants

mafernando:
will be addressed in https://github.com/projectjellyfish/api/issues/138

mafernando:
This issue has changed scope a bit given how alerts are now implemented. 

An entity (Project, Service, Organization, or Staff) can have multiple alert types associated with it.

![image](https://cloud.githubusercontent.com/assets/9356425/8479219/ea27da0c-209b-11e5-9050-2186668de700.png)

@stackus - do you think it makes sense to close this ticket?

stackus:
I have to answer that with a question. Does the request to filter out OK alerts still stand?

chriskacerguis:
@mafernando don't forget a service can have multiple current alerts

For example:

dbserver01
- CRITICAL: RAM usage at 90%
- CRITICAL: CPU usage at 99%

mafernando:
@chriskacerguis: duly noted, which of the following entities do you only want display non-OK alerts for?
- projects (the project as a whole, e.g. audit is coming)
- services (for order items in a project)
- organization (proxy for system alerts)
- staff

chriskacerguis:
all.

stackus:
Should OK alerts never be recorded then?

chriskacerguis:
No we need to see the transition to we can do reports on that

mafernando:
The non-OK alert restriction would be applied to the entry points specified on Issue https://github.com/projectjellyfish/api/issues/790.

We will need to implement those first before being able to add the non-OK restriction to Projects, Services and Staff.

mafernando:
this is now implemented for admin alerts (organizations).

the restriction has been noted on Issue https://github.com/projectjellyfish/api/issues/790 

this ticket is functionally a duplicate, so closing it.

PR #792, opened 2015-07-02
adds filter to only display latest non-ok alerts under admin alerts

mafernando:
adds filter to only display latest non-ok alerts under admin alerts

addresses Issue https://github.com/projectjellyfish/api/issues/194

### DIFF
--- a/src/client/app/states/admin/alerts/list/list.html
+++ b/src/client/app/states/admin/alerts/list/list.html
@@ -24,7 +24,7 @@
         </tr>
         </thead>
         <tbody>
-        <tr class="alerts-table__row" ng-repeat="row in vm.alerts">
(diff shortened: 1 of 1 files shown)


src/client/app/states/admin/alerts/list/list.html
```diff
@@ -24,7 +24,7 @@
         </tr>
         </thead>
         <tbody>
-        <tr class="alerts-table__row" ng-repeat="row in vm.alerts">
+        <tr class="alerts-table__row" ng-repeat="row in vm.alerts | filter:{status: '!ok'}">
```
stackus: Interesting, Angular will parse the string for '!' and evaluate it as `status != 'ok'`?

mafernando: saw it [here](http://stackoverflow.com/questions/22666366/angular-js-filter-not-equals)

PR #793, opened 2015-07-02
Fixes issue with parameters. Also removes staff_id parameter and only…

### LINKED ISSUE #789: Changes and additions to MOTD API
- [x] Endpoint to retrieve the latest MOTD
- [x] Ensure users aren't posting MOTDs as others.  We shouldn't accept `staff_id` from params and should instead assign it from `current_staff.id`.

OBCENEIKON:
… allows for one MOTD to exist.

Fixes #789


config/routes.rb
```diff
@@ -153,7 +153,17 @@
     resources :tags, only: [:index]
 
     # MOTD Routes
-    resources :motd
+    # collection do
```
stackus: comments

config/routes.rb
```diff
@@ -153,7 +153,17 @@
...
+    #   get :motd
+    #   post :motd
+    #   patch :motd
+    #   patch :motd
+    #   delete :motd
+    # end
+    get :motd, to: 'motd#index'
```
stackus: Does `resource :motd` do the same as these? Notice `resource` is singular.

OBCENEIKON: This eliminates the need to pass and id. Since there will ever only be one MOTD, we just do a first_or_initialize in the controller.

config/routes.rb
```diff
@@ -153,7 +153,10 @@
     resources :tags, only: [:index]
 
     # MOTD Routes
-    resources :motd
+    get :motd, to: 'motd#index'
```
stackus: To follow-up on my last comment. The route system has both a `resources` and a `resource` helper. The first sets up routes which will require ids, the second is for things that there is only one of and so no ids are needed.

Link again to the documentation for `resource`: http://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/Resources.html#method-i-resource

app/policies/motd_policy.rb
```diff
@@ -1,11 +1,9 @@
 class MotdPolicy < ApplicationPolicy
-  def index?
-    true
-  end
+  # def initialize(user, record); end
 
-  def show?
-    true
-  end
+  # def show?
```
stackus: comments

stackus:
@OBCENEIKON Please remove the comments and then fix the merge conflicts so this can be merged.

Issue #146, opened 2015-01-28
After project approval only admins can change budget


chriskacerguis:
After project approval on admins can change budget

mafernando:
Addressed in PR https://github.com/projectjellyfish/api/pull/796.

mafernando:
merged into master, closing.

PR #796, opened 2015-07-02
prevents non admins from editing project budget after creation.

mafernando:
if the user is not an admin and attempts to update the budget of an existing project, they are prevented from doing so by deleting the budget param from their request.
assumes that when a project id is passed as a param then an update is being handled.
addresses issue https://github.com/projectjellyfish/api/issues/146

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -86,6 +86,7 @@ class ProjectsController < ApplicationController
         project[:project_answers_attributes] = project.delete(:project_answers)
       end
     end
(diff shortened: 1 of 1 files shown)


app/controllers/projects_controller.rb
```diff
@@ -86,6 +86,7 @@ def project_params
         project[:project_answers_attributes] = project.delete(:project_answers)
       end
     end
+    @_project_params = (!current_user.admin? && !params[:id].nil?) ? @_project_params.delete(:budget) : @_project_params
```
stackus: This needs to happen within the block above. The block is being set to `@_project_params` only once, every other time it's just being recalled.

PR #819, opened 2015-07-08
Bug/791 project serializer related data

### LINKED ISSUE #791: Data returned from ProjectSerializer strips relation data
- [x] Add missing `*_serializer.rb` files for any related data in `project_serializer.rb`.
- [x] Serializer relation DSL does not accept additional parameters so things like `through:`, `class_name:`, ... can be removed

OBCENEIKON:
Fixes #791

### DIFF
--- a/app/serializers/membership_serializer.rb
+++ b/app/serializers/membership_serializer.rb
@@ -16,5 +16,9 @@
 #
 
 class MembershipSerializer < ApplicationSerializer
-  attributes :id, :created_at, :updated_at
+  attributes :id, :group_id, :project_id, :role_id, :created_at, :updated_at
+  
+  has_one :group
+  has_one :project
+  has_one :role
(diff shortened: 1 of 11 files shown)


app/serializers/membership_serializer.rb
```diff
@@ -0,0 +1,20 @@
...
+#  role_id    :integer
+#
+# Indexes
+#
+#  index_memberships_on_group_id    (group_id)
+#  index_memberships_on_project_id  (project_id)
+#
+
+class MembershipSerializer < ApplicationSerializer
+  attributes :id, :created_at, :updated_at
```
stackus: We may want to include `:group_id, :project_id, :role_id` in the `attributes` list and then also add the relationships.

```
has_one :group
has_one :project
has_one :role
```

PR #820, opened 2015-07-09
Feature/799 project apply roles

AllenBW:


src/client/app/components/project-membership-modal/_add-membership-modal.sass
```diff
@@ -0,0 +1,104 @@
++block(add-group-modal)
```
stackus: s/group/membership/

src/client/app/components/project-membership-modal/add-membership-modal.html
```diff
@@ -0,0 +1,63 @@
...
+
+    <div class="row">
+      <div class="form-field__container">
+
+        <div class="form-field" ng-disabled="!vm.currentMembership.id">
+          <div class="form-field__aside">
+            <label class="form-field__label" ng-if="!vm.currentMembership.id">Add Group</label>
+          </div>
+          <div class="form-field__body">
+            <select id="group-name" name="group" ng-model="vm.selectedGroup"
```
stackus: Should apply directly to the model with `vm.membership.group_id`.

src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      function resolveRoles() {
+        return Role.query().$promise;
+      }
+    }
+  }
+
+  /** @ngInject */
+  function AddMembershipModalController($stateParams, groups, roles, listMembership, editMembership, Membership,
```
stackus: The naming of the files and controller are confusing `ProjectMembership` and `AddMembership`. 

Also I'm not getting why you're juggling a `listMembership` and `editMembership`. You're bringing in an outside concern which is the concern of where this directive is being used.

src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js
```diff
@@ -0,0 +1,125 @@
...
+          vm.updateMembership.$update({
+              projectId: vm.currentMembership.project_id,
+              groupId: vm.currentMembership.group_id
+            },
+            updateSuccess, updateFailure);
+        } else if (!vm.currentMembership.id) {
+          if (lodash.result(lodash.find(vm.listMembership, 'group_id', vm.selectedGroup.id), 'group_id')) {
+            Toasts.error('Group selected is already associated with project.');
+          } else {
+            vm.currentMembership.group_id = $stateParams.projectId;
```
stackus: Inserted concern of `$state` and `productId`.

src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  /** @ngInject */
+  function AddMembershipModalController($stateParams, groups, roles, listMembership, editMembership, Membership,
+                                        lodash, Toasts) {
+    var vm = this;
+
+    vm.groups = groups;
+    vm.currentMembership = editMembership || new Membership();
```
stackus: The membership should be passed in already initialized. Whether it's new or existing is only a concern when you need to determine between `$save` or `$update`.

src/client/app/components/project-membership-modal/add-membership-modal.html
```diff
@@ -0,0 +1,63 @@
...
+          </div>
+        </div>
+
+        <div class="form-field">
+          <div class="form-field__aside">
+            <label class="form-field__label" ng-if="!vm.currentMembership.id">Add Group Role</label>
+            <label class="form-field__label" ng-if="vm.currentMembership.id">Update Group Role</label>
+          </div>
+          <div class="form-field__body">
+            <select id="group-role" name="role" ng-model="vm.selectedRole" class="form-field__input"
```
stackus: Apply directly to `vm.membership.role_id`.

src/client/app/components/project-membership-table/project-membership-table.directive.js
```diff
@@ -0,0 +1,68 @@
...
+      vm.deleteGroup = deleteMembership;
+      vm.openEditGroup = openEditGroup;
+      vm.rowLookup = rowLookup;
+      vm.membership = new Membership();
+
+      function activate() {
+      }
+
+      function deleteMembership(index) {
+        vm.membership.$delete({projectId: vm.memberships[index].project_id, groupId: vm.memberships[index].group_id},
```
stackus: You should be able to use the methods `$save`, `$update`, and `$delete` without having to pass in an object to work with. These methods operate on the resource object itself.

src/client/app/resources/membership.factory.js
```diff
@@ -0,0 +1,31 @@
...
+    var Membership = $resource('/api/v1/projects/:projectId/groups/:groupId',
+      {projectId: '@projectId', groupId: '@groupId'}, {
+        update: {
+          method: 'PUT',
+          isArray: false
+        }
+      });
+
+    Membership.defaults = {
+      project_id: 0,
```
stackus: Use nulls instead of 0s.

src/client/app/states/projects/details/details.state.js
```diff
@@ -104,25 +114,9 @@
...
-          vm.project.group_ids.push(vm.groupToAdd.id);
-          vm.project.$update(saveSuccess, saveFailure);
-        }
-      }
-
-      function saveSuccess() {
-        Toasts.toast('Group successfully added.');
-        vm.project.groups.push(vm.groupToAdd);
-      }
+      ProjectMembership.showModal(vm.project.memberships).then(updateMembership);
```
stackus: Why does the modal need to know about a projects list of memberships? It only needs to know about the membership resource it has to operate on.

`ProjectMembership.showModal(Membership.new()).then(updateMembership);`

src/client/app/states/projects/details/details.html
```diff
@@ -113,14 +113,18 @@
   <div class="project-details__service-buttons">
     <button type="button"
             class="btn-rounded btn-rounded--short project-details__service-button"
-            ng-click="vm.openAddGroup()">
+            ng-click="vm.openAddGroup(vm.project.memberships)">
```
stackus: `vm.openAddGroup` doesn't take any parameters. 

Suggestion 
- Rename the method to `vm.showMembershipModal`
- Have it take a single parameter
- Process the parameter such that if it's null it's assigned `Membership.new()`
- Pass the parameter to the modal.

src/client/app/states/login/login.state.js
```diff
@@ -22,6 +22,9 @@
         controller: StateController,
         controllerAs: 'vm',
         title: 'Login',
+        resolve: {
```
stackus: It was probably better to do the motd the other way. This way if it fails to resolve :boom:.

src/client/app/components/project-membership-table/project-membership-table.directive.js
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    return directive;
+
+    function link(scope, element, attrs, vm, transclude) {
+      vm.activate();
+    }
+
+    /** @ngInject */
+    function ProjectMembershipController($q, lodash, Toasts, ProjectMembership, Membership) {
```
stackus: `ProjectMembership` is perhaps poorly named? Better would be `MembershipModal`.

src/client/app/components/project-membership-modal/add-membership-modal.html
AllenBW: attempted `group.id as group.name for group in vm.groups track by group` and the same thing for role, but it borks the value in the select

src/client/app/components/project-membership-modal/_add-membership-modal.sass
AllenBW: onboard with this

src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js
AllenBW: listmembership is a copy of existing membership list (we have no endpoint to hit to get this remember) used on add group to eliminate dupes

editmembership is the row value of a membership object when editing

changing the names

src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js
AllenBW: projectId is not the same as project_id,  updateMembership reaches back to the membership factory, which accepts two parameters projectId and groupId

src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js
AllenBW: Membership (the one submitted for save or update) isn't already passed in

src/client/app/resources/membership.factory.js
```diff
@@ -0,0 +1,31 @@
+(function() {
+  'use strict';
+
+  angular.module('app.resources')
+    .factory('Membership', MembershipFactory);
+
+  /** @ngInject */
+  function MembershipFactory($resource) {
+    var Membership = $resource('/api/v1/projects/:projectId/groups/:groupId',
+      {projectId: '@projectId', groupId: '@groupId'}, {
```
stackus: Use `@project_id` and `@group_id` to access the values on the resource to have them applied to the url.

src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js
stackus: Go ahead and drop the listMembership and dupe checks for now. The server will tell us if there's a dupe and if it doesn't it needs to and we'll open a bug.

src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js
stackus: Right, but it should be. The component should get a resource and operate on that resource.

src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js
stackus: I've made a comment on the resource that will make it so you don't have to pass in an object each time.

stackus: I updated my initial comment and moved it elsewhere also. The comment for this line is that you're doing `group_id` = `projectId`.

src/client/app/components/project-membership-table/project-membership-table.directive.js
AllenBW: ? without the parameters i wasn't able to save update or delete

src/client/app/resources/membership.factory.js
AllenBW: sweet! i was worried about this. implemented.

src/client/app/states/projects/details/details.state.js
AllenBW: addressed in earlier, in order to dedupe logic of posting in the two unique places the modal is called, logic to submit was moved to the modal controller, project's list of memberships is necessary to avoid adding two identical groups to a project.  the scoping also provides a handy means of modifying the view without any additional effort

src/client/app/resources/membership.factory.js
```diff
@@ -0,0 +1,31 @@
...
+
+    Membership.defaults = {
+      project_id: 0,
+      group_id: 0,
+      role_id: 0
+    };
+
+    Membership.new = newMembership;
+
+    function newMembership() {
```
stackus: Realized something that a lot `.new()` functions could use is to accept an object that gets merged with the copy data.
```
function newMembership(data) {
  return new Membership(angular.extend({}, Membership.defaults, data));
}
```

src/client/app/states/projects/details/details.html
AllenBW: good catch, removed parameter. modified name.  need the two parameters for list deduplication

src/client/app/states/login/login.state.js
AllenBW: funny you should mention this... i feel the same way, but struggled making the resource available to the controller when it was acquired by the suggested... 
```  function StateController(Motd) {
    var vm = this;

    vm.title = 'Login';
    vm.resolveMotd = resolveMotd;
    activate();

    function activate() {
      vm.resolveMotd();
    }

    function resolveMotd() {
      Motd.get().$promise.then(assignResolved());

      function assignResolved(result) {
        vm.motd = result;
      }
    }
  }```

File: src/client/app/components/project-membership-table/project-membership-table.directive.js (hunk 12 above)
Comment (AllenBW) [inline src/client/app/components/project-membership-table/project-membership-table.directive.js]: disagree, this is the directive for the membership table of the project details page (house baratheon, son of the andals n' stuff)  has nothing to do with the modal, thems concerns are seperated

File: src/client/app/components/project-membership-modal/add-membership-modal-service.factory.js [hunk 15]
```diff
@@ -0,0 +1,114 @@
...
+    function activate() {
+    }
+
+    function onSubmit() {
+      vm.showValidationMessages = true;
+
+      if (vm.form.$valid) {
+        if (vm.currentMembership.id) {
+          vm.currentMembership.role_id = vm.selectedRole.id;
+          vm.updateMembership = new Membership();
```
stackus: Rework so you don't have to use any extra membership objects. A single one on the vm should be enough.

src/client/app/components/project-membership-table/project-membership-table.directive.js
stackus: How would another developer know which is a Resource and which is a Modal service when faced with `ProjectMembership` and `Membership`?

PR #822, opened 2015-07-09
Added rufus task for update_budgets.

### LINKED ISSUE #441: Move Rake tasks to rufus
So, we need to move the rake tasks that we have to use the new rufus system.  From the install guide

The following rake command needs to be put into Rufus.

````
# Updates the budgets for projects
rake upkeep:update_budgets
````

We don't care about the other 2 jobs at the moment.

OBCENEIKON:
Fixes #441

### DIFF
--- a/config/initializers/task_scheduler.rb
+++ b/config/initializers/task_scheduler.rb
@@ -10,3 +10,7 @@ scheduler = Rufus::Scheduler.new
 scheduler.every '24h' do
   Rake::Task['upkeep:prune_alerts']
 end
+
+scheduler.every '15m' do
+  Rake::Task['upkeep:update_budgets']
+end


mafernando:
I noticed that Project has drifted from the original implementation of `update_budgets` and no longer works when you run it. 

In particular, [`approved`](https://github.com/projectjellyfish/api/blob/master/lib/tasks/upkeep.rake#L64) no longer exists on the Project model. It is replaced by [approvals](https://github.com/projectjellyfish/api/blob/master/db/schema.rb#L362).

This could be a quick fix by replacing [`Project.where(approved: true)`](https://github.com/projectjellyfish/api/blob/master/lib/tasks/upkeep.rake#L64) with `Project.where('approval <> 0')`, but approvals appears to be setup weird. 

It is defined as both an [approval relationship](https://github.com/projectjellyfish/api/blob/master/app/models/project.rb#L44) on Project and an [approval enum](https://github.com/projectjellyfish/api/blob/master/app/models/project.rb#L52).

I think we need to move the [approval enum](https://github.com/projectjellyfish/api/blob/master/app/models/project.rb#L52) to the Approval model to be used in the [`approved`](https://github.com/projectjellyfish/api/blob/master/db/schema.rb#L48) column.

Then in upkeep_budgets do something like `Project.where(id: Approval.where(approved: true).select('id'))` could be used to get approved projects to compute budget for.

PR #828, opened 2015-07-10
Bug/742 edit project questions

AllenBW:
This fixes the issue of not being able to edit project questions, BUT this doesn't fix the issue of how project question options are displayed in edit or on a project details page (new issues incoming)


stackus:
Odd question. This is complete? So used to seeing massive PRs lately.

Issue #831, opened 2015-07-10
Rufus crash


stackus:
Looks like the task Rufus is supposed to execute is missing or incorrectly named.

```
{ 70148382805380 rufus-scheduler intercepted an error:
  70148382805380   job:
  70148382805380     Rufus::Scheduler::EveryJob "15m" {}
  70148382805380   error:
  70148382805380     70148382805380
  70148382805380     RuntimeError
  70148382805380     Don't know how to build task 'upkeep:update_budgets'
  70148382805380       /Users/stack/.rbenv/versions/2.2.2/lib/ruby/2.2.0/rake/task_manager.rb:62:in `[]'
  70148382805380       /Users/stack/.rbenv/versions/2.2.2/lib/ruby/2.2.0/rake/task.rb:353:in `[]'
  70148382805380       /Users/stack/source/bah/api/config/initializers/task_scheduler.rb:15:in `block in <top (required)>'
  70148382805380       /Users/stack/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rufus-scheduler-3.1.3/lib/rufus/scheduler/jobs.rb:234:in `call'
  70148382805380       /Users/stack/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rufus-scheduler-3.1.3/lib/rufus/scheduler/jobs.rb:234:in `do_call'
  70148382805380       /Users/stack/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rufus-scheduler-3.1.3/lib/rufus/scheduler/jobs.rb:258:in `do_trigger'
  70148382805380       /Users/stack/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rufus-scheduler-3.1.3/lib/rufus/scheduler/jobs.rb:300:in `block (3 levels) in start_work_thread'
  70148382805380       /Users/stack/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rufus-scheduler-3.1.3/lib/rufus/scheduler/jobs.rb:303:in `call'
  70148382805380       /Users/stack/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rufus-scheduler-3.1.3/lib/rufus/scheduler/jobs.rb:303:in `block (2 levels) in start_work_thread'
[...]

PR #832, opened 2015-07-10
Enhancement/441 move tasks to rufus

### LINKED ISSUE #441: Move Rake tasks to rufus

OBCENEIKON:
Fixes #441 and #831

### DIFF
--- a/lib/tasks/upkeep.rake
+++ b/lib/tasks/upkeep.rake
@@ -63,7 +63,6 @@ namespace :upkeep do
 
   desc 'Update Remaining Project Budgets'
   task update_budgets: :environment do
-
     options = {}
     OptionParser.new do |opts|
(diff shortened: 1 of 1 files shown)


stackus:
Rake Tasks have their own method of taking parameters. See the top answer here.

http://stackoverflow.com/questions/825748/how-do-i-pass-command-line-arguments-to-a-rake-task

The reason this should be changed to the rake way of doing things is because developers who know rake will be surprised by the change of how they are supposed to pass parameters into a rake task.

PR #839, opened 2015-07-13
Use console syntax to execute scheduled rake tasks.

mafernando:
Addresses issue https://github.com/projectjellyfish/api/issues/831.

Use console syntax to execute scheduled rake tasks.

Stackoverflow [suggested](http://stackoverflow.com/questions/8049361/rufus-scheduler-fails-with-rake-tasks-dont-know-how-to-build-task) this approach and it seemed to work.

### DIFF
--- a/config/initializers/task_scheduler.rb
+++ b/config/initializers/task_scheduler.rb
@@ -8,9 +8,9 @@ require 'rake'
 scheduler = Rufus::Scheduler.new
 
 scheduler.every '24h' do
-  system('rake upkeep:prune_alerts')
+  system('bundle exec rake upkeep:prune_alerts')
 end
 
 scheduler.every '15m' do
-  system('rake upkeep:update_budgets')
+  system('bundle exec rake upkeep:update_budgets')
 end
--- a/config/initializers/task_scheduler.rb
+++ b/config/initializers/task_scheduler.rb
@@ -8,9 +8,9 @@ require 'rake'
 scheduler = Rufus::Scheduler.new
 
 scheduler.every '24h' do
-  Rake::Task['upkeep:prune_alerts']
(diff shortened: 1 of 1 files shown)


config/initializers/task_scheduler.rb
```diff
@@ -8,9 +8,9 @@
 scheduler = Rufus::Scheduler.new
 
 scheduler.every '24h' do
-  Rake::Task['upkeep:prune_alerts']
+  system('rake upkeep:prune_alerts')
```
stackus: I think prefixing the command with `bundle exec` might help. Some installs might be grumpy without it.

mafernando: Absolutely agree with you! Done.

config/initializers/task_scheduler.rb
```diff
@@ -8,9 +8,9 @@
 scheduler = Rufus::Scheduler.new
 
 scheduler.every '24h' do
-  Rake::Task['upkeep:prune_alerts']
+  system('bundle exec rake upkeep:prune_alerts')
```
chriskacerguis: Not sure here, but I'm wondering how secure doing this is?  @mafernando thoughts?

mafernando: Hmm, I poked around a little but and the primary thing that I found to be of concern is the potential for command injection. In particular, see Exapmle 4 from [this](https://www.owasp.org/index.php/Command_Injection) page. Since we're not passing any parameters, the only issue I can think of would be if someone overwrote the upkeep methods - but that could happen regardless since this is an open source project. @stackus - any other perspectives on this?

chriskacerguis: Makes sense.  :+1:

PR #842, opened 2015-07-14
Bug/829 Project Edit, broken project options questions

AllenBW:
Does not yet address the project question edit state.

### DIFF
--- a/src/client/app/components/project-form/project-form.directive.js
+++ b/src/client/app/components/project-form/project-form.directive.js
@@ -79,21 +79,10 @@
               'url', 'state', 'state_ok', 'problem_count', 'account_number', 'resources', 'icon', 'status', 'users',
               'order_history', 'cc', 'staff_id', 'approved', 'project_answers');
             if (angular.isDefined(vm.project.project_answers) && (vm.project.project_answers.length > 0)) {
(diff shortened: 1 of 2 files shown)


src/client/app/components/project-form/project-form.directive.js
```diff
@@ -79,21 +79,9 @@
...
-
-                  return pas;
-                }, []);
-            }
-
-            for (var prop in vm.filteredProject) {
-              if (vm.filteredProject[prop] === null) {
-                delete vm.filteredProject[prop];
-              }
+              vm.filteredProject.project_answers = lodash.reduce(
```
stackus: `lodash.map` is better suited for use here.

src/client/app/components/project-form/project-form.directive.js
```diff
@@ -104,6 +92,11 @@
           }
         }
 
+        function projectAnswerReduction(item) {
+          return lodash.merge({'project_question_id': item.project_question.id}, {'id': item.id},
```
stackus: Using `lodash.merge` is unnecessary. 

`return {id: item.id, project_question_id: item.project_question.id, answer: item.answer};`

PR #845, opened 2015-07-16
Bug/827 product type provisioning answers

AllenBW:


src/client/app/components/project-form/project-form-options.html
```diff
@@ -0,0 +1,91 @@
+<div class="project-field" ng-repeat="answer in vm.projectOptions">
+  <div class="project-field__aside">
+    <label for="{{answer.project_question.id}}"
+           class="project-field__label">{{answer.project_question.question}}</label>
+  </div>
+                <span ng-switch on="answer.project_question.field_type">
```
stackus: `span`s shouldn't contain block elements.

PR #848, opened 2015-07-17
addresses all three asks of issue 800.

AllenBW:


src/client/app/components/monthly-price/monthly-price.directive.js
```diff
@@ -0,0 +1,45 @@
...
+
+  angular.module('app.components')
+    .directive('computedMonthlyPrice', ComputedMonthlyPriceDirective);
+
+  /** @ngInject */
+  function ComputedMonthlyPriceDirective() {
+    var directive = {
+      restrict: 'AE',
+      scope: {
+        hourlyPrice: '=',
```
stackus: Accept an object that you can read the properties from. Reason to do that is because if the business logic for computing the total has to change we'll have to update every use of the directive. Accepting quantity as a multiplier is fine.

src/client/app/components/monthly-price/monthly-price.directive.js
```diff
@@ -0,0 +1,45 @@
...
+        monthlyPrice: '=',
+        quantity: '=?'
+      },
+      link: link,
+      templateUrl: 'app/components/monthly-price/monthly-price.html',
```
stackus: Inline the template instead. To do that use `template: ""`, so `template` instead of `templateUrl`.

src/client/app/components/monthly-price/monthly-price.directive.js
```diff
@@ -0,0 +1,45 @@
...
+      var vm = this;
+
+      vm.activate = activate;
+      vm.computeMonthlyTotal = computeMonthlyTotal;
+
+      function activate() {
+        vm.quantity = vm.quantity || 1;
+      }
+
+      function computeMonthlyTotal(hourlyprice, monthlyprice) {
```
stackus: This should be parameter-less. The directive is meant to be self contained and nothing will ever be able to pass in different values anyhow.

src/client/app/components/compare/compare-modal.html
```diff
@@ -21,7 +21,7 @@
         <div class="compare-modal__cell compare-modal__cell--value"
              ng-repeat="value in ::row.values track by $index"
              ng-class="{'compare-modal__cell--even': $even}">
-          {{ ::value }}
+          {{ ::value | number:2 }}
```
stackus: Adding a number filter will truncate hourly prices to 0.00, and will fail on other text properties. Remember that this compare dialog is supposed to show other properties and it's a bug (reported elsewhere) that it is not.

src/client/app/components/cart/cart-modal.html
```diff
@@ -27,7 +28,12 @@
       <td class="cart-table__cell cart-table__cell--product">
         <product-description product="line.product" link-to="products.details({productId: {{line.product.id}}})"></product-description>
       </td>
-      <td class="cart-table__cell cart-table__cell--price">{{ line.price | currency }}</td>
+      <td class="cart-table__cell cart-table__cell--price">{{ line.product.monthly_price | currency }}</td>
```
stackus: This should be `setup_price` I think.

PR #857, opened 2015-07-21
Enhancement/801 wizard back

### LINKED ISSUE #801: Users cannot go back to previous questions in the wizard
The wizard needs a back feature so that users can revisit a question and change their answer.

- [x] Add a back button to the wizard

AllenBW:
closes #801

### DIFF
--- a/src/client/app/components/wizard/_wizard-modal.sass
+++ b/src/client/app/components/wizard/_wizard-modal.sass
@@ -54,6 +54,17 @@
     &:last-child
       margin-bottom: 0
 
+    +modifier(selected)
+      &:before
+        content: ""
+        height: 0
+        left: 250px
+        position: absolute
+        width: 0
+        border: 10px solid transparent
+        border-left-color: $app-color-dark-green
+
+
   +element(footer)
     @extend .modal-footer
     text-align: center
--- a/src/client/app/components/wizard/_wizard-modal.sass
+++ b/src/client/app/components/wizard/_wizard-modal.sass
@@ -50,6 +50,7 @@
     display: block
     min-width: 200px
     margin: 0 auto 20px
+    position: relative
 
     &:last-child
       margin-bottom: 0
@@ -58,7 +59,7 @@
       &:before
         content: ""
         height: 0
-        left: 250px
+        left: -20px
         position: absolute
         width: 0
         border: 10px solid transparent
--- a/src/client/app/components/wizard/wizard-modal.html
+++ b/src/client/app/components/wizard/wizard-modal.html
@@ -15,7 +15,7 @@
       <button type="button"
               class="btn-rounded btn-rounded--capitalize wizard-modal__button"
               ng-repeat="answer in vm.question.wizard_answers track by answer.id"
-              ng-class="{'wizard-modal__button--selected': vm.answeredQuestions[vm.question.id].id === answer.id}"
+              ng-class="{'wizard-modal__button--selected': vm.answeredQuestions[vm.questionPointer].id === answer.id}"
               ng-click="vm.answerWith($index)">
         {{ ::answer.text }}
       </button>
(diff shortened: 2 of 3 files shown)


src/client/app/components/wizard/_wizard-modal.sass
```diff
@@ -54,6 +54,17 @@
     &:last-child
       margin-bottom: 0
 
+    +modifier(selected)
+      &:before
+        content: ""
+        height: 0
+        left: 250px
```
stackus: If an answer happens to be wider than the `min-width` this could cause the indicator to be out of position. You can use `position: relative` on the `button` or `answers` elements to avoid having to using such a large value.

src/client/app/components/wizard/wizard-modal.html
```diff
@@ -27,5 +29,14 @@
   </div>
 </div>
 <div class="wizard-modal__footer">
-  <button type="button" class="btn-rounded btn-rounded--gray wizard-modal__ok" ng-click="$dismiss()">Cancel</button>
+  <button type="button"
+          ng-disabled="!vm.questionsArchive.length || vm.state === 'intro' || vm.state === 'complete'"
```
stackus: `|| vm.state === 'intro' || vm.state === 'complete'` can shortened to be `|| vm.state != 'wizard'`

src/client/app/components/wizard/wizard-service.factory.js
```diff
@@ -51,22 +54,35 @@
...
+        lodash.forEach(vm.answeredQuestions, parseQuestionAnswers);
         vm.state = 'complete';
       }
     }
+
+    function parseQuestionAnswers(item) {
+      vm.tags = lodash.without(lodash.union(vm.tags, item.tags_to_add), item.tags_to_remove);
+    }
+
+    function previousQuestions() {
```
stackus: Suggest that instead of managing two lists of questions and juggling the current question between them that you instead use a pointer to the current question instead. The pointer value should be initialized to 0, and have a max value equal to the length of questions minus one. 

Also keep track of the index in `answerWith` as the answer. The length of the answers list should match the number of questions the user has answered with it's max length being equal to the length of the questions list.

AllenBW: to the second point, what you mention is already being done, the key of answerWith serves as the index and is compared with the current vm.question id # to know when to disable Next

to the first point, shift was being used previously, was looking to follow the style of what was in place, a pointer rather than two lists for performance reasons?

stackus: It kinda is, but it's doing it by association. Line 63 is certainly doing more than recording `index` into an array.

The code before was throwing questions away and using that to determine if there were anymore. It's not going to result in better performance to use a pointer and a simple array of integers but when this code is revisited it will be more clear to whomever is working on it what is happening.

AllenBW: fair enough, it did require a bit of noodling to figure out what exactly was going on initially....

src/client/app/components/wizard/wizard-service.factory.js
```diff
@@ -39,10 +39,14 @@
     vm.state = 'intro';
     vm.questions = questions;
     vm.question = null;
-    vm.tags = [];
+    vm.questionPointer = 0;
+    vm.questionsEnd = lodash.size(vm.questions) - 1;
```
stackus: lodash isn't needed here. `vm.questions.length - 1`

src/client/app/components/wizard/wizard-service.factory.js
```diff
@@ -50,23 +54,35 @@
...
       vm.state = 'wizard';
     }
 
     function answerWith(index) {
-      var answer;
-
       if (0 <= index) {
-        answer = vm.question.wizard_answers[index];
-        vm.tags = lodash.without(lodash.union(vm.tags, answer.tags_to_add), answer.tags_to_remove);
+        vm.answeredQuestions.push(vm.question.wizard_answers[index]);
```
stackus: If they go back and change their answer this would push a new answer onto the stack.

src/client/app/components/wizard/wizard-service.factory.js
```diff
@@ -39,10 +39,14 @@
...
     vm.questions = questions;
     vm.question = null;
-    vm.tags = [];
+    vm.questionPointer = 0;
+    vm.questionsEnd = lodash.size(vm.questions) - 1;
+    vm.answeredQuestions = [];
 
     vm.startWizard = startWizard;
     vm.answerWith = answerWith;
+    vm.previousQuestions = previousQuestions;
```
stackus: These two variables, next and previous questions, are no longer needed right?

PR #875, opened 2015-07-27
password optional on user edit

### LINKED ISSUE #871: Cannot edit a user without being forced to change their password
When editing a user the password fields should be optional.

AllenBW:
closes #871

### DIFF
--- a/src/client/app/components/user-form/user-form.html
+++ b/src/client/app/components/user-form/user-form.html
@@ -67,11 +67,15 @@
                     <label class="user-field__label">Password</label>
                 </div>
                 <div class="user-field__body" ng-class="{'has-error': vm.hasErrors('password')}">
-                    <input id="user-password" name="password" type="password"
+                    <input id="user-password"
+                           name="password"
+                           type="password"
                            class="user-field__input"
                            ng-model="firstpass"
                            ng-minlength="8"
-                           placeholder="Must be 8 or more characters."  required />
+                           placeholder="Must be 8 or more characters."
+                           ng-required="!vm.editing"
+                          />
                     <div class="user-field__feedback" ng-messages="vm.form.password.$error">
                         <div ng-message="required">A password is required.</div>
                         <div ng-message="minlength">Password must be at least 8 Characters.</div>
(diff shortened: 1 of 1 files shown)


src/client/app/components/user-form/user-form.html
```diff
@@ -67,11 +67,15 @@
...
-                    <input id="user-password" name="password" type="password"
+                    <input id="user-password"
+                           name="password"
+                           type="password"
                            class="user-field__input"
                            ng-model="firstpass"
                            ng-minlength="8"
-                           placeholder="Must be 8 or more characters."  required />
+                           placeholder="Must be 8 or more characters."
+                           ng-required="!vm.editing"
```
stackus: How is `vm.editing` being set?

src/client/app/components/user-form/user-form.html
```diff
@@ -84,9 +88,13 @@
...
-                    <input id="user-confirmpassword" ng-model="vm.userToEdit.password"
-                           name="confirmpassword" type="password"
-                           class="user-field__input" placeholder=""  required />
+                    <input id="user-confirmpassword"
+                           ng-model="vm.userToEdit.password"
+                           name="confirmpassword"
+                           type="password"
+                           class="user-field__input"
+                           placeholder=""
+                           ng-required="firstpass"/>
```
stackus: Where is `firstpass` coming from?

AllenBW: already existed, used for matching password confirmation, didn't touch, but
it doesn't work at the moment

src/client/app/components/user-form/user-form.html
```diff
@@ -67,11 +67,15 @@
                     <label class="user-field__label">Password</label>
                 </div>
                 <div class="user-field__body" ng-class="{'has-error': vm.hasErrors('password')}">
-                    <input id="user-password" name="password" type="password"
+                    <input id="user-password"
+                           name="password"
+                           type="password"
                            class="user-field__input"
                            ng-model="firstpass"
```
stackus: Never setting the password onto the model means it'll never be sent to the server.

src/client/app/components/user-form/user-form.html
AllenBW: directive parameter based on presence or absence of item id, going to
author an issue to clean up this form, that'll be one of the things to fade
away

src/client/app/components/user-form/user-form.html
AllenBW: quite true, its set in the confirmation

PR #879, opened 2015-07-27
switched checkboxes for radio buttons

### LINKED ISSUE #869: Styling on the Roles pages are not inline with the rest of UX
The roles pages need their styles fixed.

- [x] List page table is missing border
- [x] List page actions are missing the 'Action' heading
- [x] List page actions are not spaced correctly
- [x] List page table applies a hover effect as you move over rows
- [x] Role form checkboxes are not using the correct native repalcement
- [x] Role form checkboxes are backwards (Should be box then label)

AllenBW:
closes #869

### DIFF
--- a/src/client/app/components/role-form/role-form.html
+++ b/src/client/app/components/role-form/role-form.html
@@ -43,25 +43,23 @@
         <div class="role-field__aside">
           <label class="role-field__label">{{permission}}</label>
         </div>
-        <div class="role-field__body--checkbox">
-          <label>Read
-            <input id="approvals-read"
-                   name="approvals-read"
-                   type="checkbox"
+        <div class="role-field__radio">
+
+            <input id="{{permission}}+{{value}}+'read'"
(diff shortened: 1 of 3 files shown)


src/client/app/components/role-form/role-form.html
```diff
@@ -43,25 +43,23 @@
...
           <label class="role-field__label">{{permission}}</label>
         </div>
-        <div class="role-field__body--checkbox">
-          <label>Read
-            <input id="approvals-read"
-                   name="approvals-read"
-                   type="checkbox"
+        <div class="role-field__radio">
+
+            <input id="{{permission}}+{{value}}+'read'"
```
stackus: These lines might be better optimized to `{{::[permission, value, '**PERM**'].join('-')}}`. Should be able to drop around 13+ watches.

AllenBW: oh nice, great call! doing doing...

PR #885, opened 2015-07-29
Enhancement/866 back button

### LINKED ISSUE #866: Pages need back functionality
Sections or places that you 'drill down into' should have a back link. An example of a page like this would be the services details page.

- [x] Enumerate the pages that need back links
- [x] Create `back-link` directive
- [x] Develop UX for `back-link`, this includes determining placement on pages
- [x] Add `back-link` to pages

AllenBW:
closes #866


src/client/app/components/back-link/back-link.directive.js
```diff
@@ -0,0 +1,79 @@
...
+      previous = lastPrevious;
+      lastPrevious = null;
+    });
+
+    $rootScope.$on('$stateChangeSuccess', function() {
+      lastPrevious = null;
+    });
+
+    var $previousState = {
+      get: function() {
```
stackus: Inlined functions.

src/client/app/components/back-link/back-link.directive.js
```diff
@@ -0,0 +1,79 @@
...
+
+  angular.module('app.components')
+    .service('$previousState', PreviousStateService);
+
+  /** @ngInject */
+  function PreviousStateService($rootScope, $state) {
+    var previous = null;
+    var lastPrevious = null;
+
+    $rootScope.$on('$stateChangeStart', function(evt, toState, toStateParams, fromState, fromStateParams) {
```
stackus: Inlined functions everywhere.

src/client/app/components/back-link/back-link.html
```diff
@@ -0,0 +1 @@
+<button class="nav-back-link__button btn-link  pull-right" ng-click="vm.stateTransition()">Back</button>
```
stackus: This should probably be inlined into the directive since it is so short.

Instead of putting `btn-link` and `pull-right` in the class list you should extend them in the block styles.

src/client/app/states/order-history/details/details.html
```diff
@@ -2,6 +2,7 @@
   <div class="detail-heading">
     <span class="content-header__title">Order Details</span>
   </div>
+  <back-link></back-link>
```
stackus: Use a div with `content-header__aside`, and wrap the back-link. If you do this you can eliminate the `pull-right` class because the aside aligns things to the right.

src/client/app/components/back-link/back-link.directive.js
AllenBW: lololololol on it.

src/client/app/components/back-link/back-link.directive.js
```diff
@@ -0,0 +1,89 @@
...
+    function goFunction() {
+      var to = $previousState.get();
+
+      return $state.go(to.state, to.params);
+    }
+
+    return $previousState;
+  }
+
+  angular.module('app.components').run(['$previousState', function($previousState) {
```
stackus: This and the next line look incorrect.

AllenBW: 50 and 51? oh! The inline function?

stackus: - It is doing injection with an array
- The function is actually empty
- Why is an empty function being `.run()`?

src/client/app/components/back-link/back-link.directive.js
```diff
@@ -0,0 +1,90 @@
...
+  angular.module('app.components')
+    .directive('backLink', BackLinkDirective);
+
+  /** @ngInject */
+  function BackLinkDirective() {
+    var directive = {
+      restrict: 'AE',
+      scope: {},
+      link: link,
+      template: '<button class="nav-back-link__button btn-link" ng-click="vm.stateTransition()">Back</button>',
```
stackus: Is there a missing sass file for this? Was it checked in before?

AllenBW: there was, but it isn't necessary, no custom styling required for the
button to manifest as desired

PR #897, opened 2015-09-09
Feature/add jellyfish notifications

mafernando:
Adds wisper and jellyfish notification gems


app/controllers/orders_controller.rb
```diff
@@ -118,4 +119,10 @@ def load_order_items
     @order = Order.find params.require(:id)
     @order_items = query_with OrderItem.where(order_id: @order.id), :includes, :pagination
   end
+
+  def publish_order_create
+    recipients = current_user.email
+    orders_url = "#{request.protocol}#{request.host_with_port}/order-history"
+    publish(:publish_order_create, @order, recipients, orders_url)
```
stackus: This is forcing any listener of `:publish_order_create` to accept a list of email addresses. If a listener has no intention of doing anything with emails it's an odd parameter to pass along into it.

Also passing along a url isn't needed either. Mailers can be configured in an initializer to set the default host to use in emails. Finally, Rails provides helpers such as `orders_url` when you declare routes in the routes file. Most routes will have a haler defined. To see them run `rake routes`, the first column is what you'd used prefixed onto either `_path` or `_url`.

app/controllers/project_approvals_controller.rb
```diff
@@ -41,4 +43,10 @@ def perform_in_transaction(project)
   def project
     @_project ||= Project.find(params[:project_id]).tap { |proj| authorize(proj) }
   end
+
+  def publish_project_approval_update
+    recipients = Staff.admin.pluck(:email).join(', ')
+    project_url = "#{request.protocol}#{request.host_with_port}/projects/#{project.id}"
+    publish(:publish_project_approval_update, project, recipients, project_url) if project.approval == 'approved'
```
stackus: Same comment as orders.

app/controllers/projects_controller.rb
```diff
@@ -107,4 +109,12 @@ def build_empty_answers_to_questions(project)
   def project
     @_project ||= Project.find(params[:id])
   end
+
+  private
+
+  def publish_project_create
+    recipients = { project_approvers: Staff.admin.pluck(:email).join(', '), project_creator: current_user.email }
+    projects_url = "#{request.protocol}#{request.host_with_port}/projects"
+    publish(:publish_project_create, OpenStruct.new(body:  response.body), recipients, projects_url)
```
stackus: Same email and url comments as before. I'd suggest instead of doing this as `after_action`s that you instead do the `publish` from within the action themselves. This goes for all three.

It'd help here double so because then you'll be able to actually have the project object that was created.

app/controllers/orders_controller.rb
mafernando: Should be addressed by https://github.com/projectjellyfish/api/commit/30837e742982258657bdd07f75a15a453ea4cc5a.

app/controllers/project_approvals_controller.rb
mafernando: Should be addressed by https://github.com/projectjellyfish/api/commit/68adb74b60a8896c0c26f9f23cc80ec57502ffcf.

app/controllers/projects_controller.rb
mafernando: Should be addressed by https://github.com/projectjellyfish/api/commit/830ffd63056a80e739537cdf17a6fe13d28c5c06.

app/controllers/projects_controller.rb
```diff
@@ -51,6 +52,7 @@ def create
     authorize Project
     group_ids = current_user.admin? ? params[:group_ids] : params.require(:group_ids)
     project = Project.create project_params.merge(group_ids: group_ids)
+    publish(:publish_project_create, project, current_user) if project.valid?
```
stackus: Use `.persisted?` instead.

PR #899, opened 2015-09-11
Modules Refactoring

stackus:
Modules can now be build on top of Jellyfish::Extension, they can require a minimum version of Jellyfish, can require a minimum version of other extensions.

- Products belong to a base provisioner
- Products are built by using custom classes for ProductType, and Service
- Provisioners are registered with custom classes of RegisteredProvisioner
- Modules can provide new UX (early work, still much to do)
- Admin features have been broken into Manager and Administrator access
- Settings ability to manage application settings inside of the app, including settings provided by modules
- Multiple providers of each type can be created
- Products can require user input at time of order
- Cart removed, ordering happens directly from the marketplace.
- UX code reduced (CSS reduced 24%, JS reduced 18%)
- Standardized a lot of the CSS classes
- Moved dashboard to '/' and login to '/login'
- Fixed a long existing bug which caused API requests to stack up
- UX updated to Angular 1.4.x
- Moved many forms to use Angular Formly object definitions

And a mountain of other small changes.

@mafernando @AllenBW @sonejah21 @chriskacerguis


chriskacerguis:
too much has changed for review.  so, I'll just give this a :+1:

stackus:
Going to fix some CodeClimate issues before merging.

PR #903, opened 2015-09-16
Fix issues with creating a new order

stackus:
- Fix creating an order with no service answers
- Fix authorizing creating an order
- Fix double render issue with UnauthorizedAccessDetection

@mafernando


mafernando:
Looks good to me, I was able to add a new order to a project. :+1:

PR #906, opened 2015-09-18
Feature/fix jel 43

mafernando:
This PR fixes [JEL-45](https://boozallen.atlassian.net/browse/JEL-43) by addressing issues with project.monthly_spend being broken in RSpec.

### DIFF
--- a/app/use_cases/create_service_order.rb
+++ b/app/use_cases/create_service_order.rb
@@ -42,16 +42,14 @@ class CreateServiceOrder
       fail UnapprovedProject, "Project '#{project.name}' has not been approved."
     end
 
-    unless params[:service]['name'].present?
+    unless params[:name].present?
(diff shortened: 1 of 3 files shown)


app/use_cases/create_service_order.rb
```diff
@@ -63,7 +61,7 @@ def build_order
       hourly_price: product.hourly_price,
       monthly_price: product.monthly_price,
       service: service
-    )
+    ).except(:name)
```
stackus: Why did you remove the services bit above only to exclude it here.

This needs to be undone because UX is going to be sending in a services object with a name value.

mafernando: Should be addressed in PR https://github.com/projectjellyfish/api/pull/908

PR #908, opened 2015-09-18
Rollback use case changes per Stack's feedback

mafernando:
Corrected project spec to pass a service hash to the CreateServiceOrder
object as it expects to receive in the OrdersController. Rolled back
changes to CreateServiceOrder object per Stack' feedback to previous
implementation.

Resolves: JEL-43


stackus:
:+1:

PR #909, opened 2015-09-18
Return 404 when delete staff not found on group

mafernando:
Previous implementation was returning `head :ok` for all delete requests
which caused delete on nonexistent staff spec to fail. Added a rescue to
return 404 on attempts to delete nonexistent staff from controller.

Resolves: JEL-42


app/controllers/associations_controller.rb
```diff
@@ -18,8 +18,12 @@ def create
   param_group :association
   def destroy
     authorize group
-    group.staff.delete staff
-    head :ok
+    begin
```
stackus: Don't use exceptions this way.

There is already a handler for these exceptions, take out the `begin`, `rescue` and `end` and let the exception get caught by the handler.

By handler I'm talking about the `rescue_with` handler defined in either `ApplicationController` or a concern.

mafernando: Hmm, I looked in the ApplicationController and only saw [these](https://github.com/projectjellyfish/api/blob/feature/add-jellyfish-docker/app/controllers/application_controller.rb#L7-L12) errors supported `ActiveRecord::RecordNotFound` is not included there. 

Should I create a new concern for `ActiveRecord::RecordNotFound`?

mafernando: Actually - I see that `MissingRecordDetection` is the concern for `ActiveRecord::RecordNotFound`. Will try to refactor this so that gets invoked correctly.

mafernando: Should be addressed by e5b03552f60ba936f8400c2a29e094a030ae7b73

PR #912, opened 2015-09-22
Added in a dynamic "jellyfish_version" to the footer

sonejah21:
Edited the directive and the html

### DIFF
--- a/src/client/app/components/footer/footer.directive.js
+++ b/src/client/app/components/footer/footer.directive.js
@@ -8,9 +8,6 @@
   function FooterDirective() {
     var directive = {
       restrict: 'AE',
-      scope: {
-        jellyfish_version: '='
-      },
       replace: true,
       link: link,
       templateUrl: 'app/components/footer/footer.html',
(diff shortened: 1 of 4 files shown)


src/client/app/components/footer/footer.directive.js
```diff
@@ -8,9 +8,41 @@
...
     var directive = {
       restrict: 'AE',
-      templateUrl: 'app/components/footer/footer.html'
+      scope: {
+        jellyfish_version: '='
+      },
+      replace: true,
+      link: link,
+      templateUrl: 'app/components/footer/footer.html',
+      controller: SulyController,
```
mafernando: Hmm, maybe a more generic name for the controller here :)

src/client/app/components/footer/footer.directive.js
```diff
@@ -8,9 +8,41 @@
...
     };
 
     return directive;
+
+    function link(scope, element, attrs, vm, transclude, Version) {
+      vm.activate();
+    }
+
+    /** @ngInject */
+    function VersionController($http) {
```
stackus: `FooterController`

src/client/app/components/footer/footer.directive.js
```diff
@@ -8,9 +8,41 @@
...
+      var vm = this;
+
+      vm.activate = activate;
+
+      function activate() {
+        initVersion();
+      }
+
+      function initVersion() {
+        $http.get('/api/v1/version/')
```
stackus: Create a resource factory for this instead of using `$http` in the states. This will also keep paths isolated in that directory and not spread into states.

src/client/app/components/footer/footer.directive.js
```diff
@@ -8,9 +8,41 @@
   function FooterDirective() {
     var directive = {
       restrict: 'AE',
-      templateUrl: 'app/components/footer/footer.html'
+      scope: {
```
stackus: `scope` should be removed. It is not being used.

PR #913, opened 2015-09-23
Editing the version in footer using new VersionFactory

sonejah21:


src/client/app/components/footer/footer.directive.js
```diff
@@ -8,9 +8,37 @@
...
+      function activate() {
+        updateVersion();
+      }
+
+      function updateVersion() {
+        Version.get().$promise.then(handleResults);
+
+        function handleResults(version) {
+          vm.version = version.jellyfish_version;
+          console.log(version.jellyfish_version);
```
stackus: logging

src/client/app/components/footer/footer.directive.js
```diff
@@ -8,9 +8,37 @@
...
+      link: link,
+      templateUrl: 'app/components/footer/footer.html',
+      controller: FooterController,
+      controllerAs: 'vm',
+      bindToController: true
     };
 
     return directive;
+
+    function link(scope, element, attrs, vm, transclude, Version) {
```
stackus: Version cannot be made available here. This function isn't injected.

src/client/app/components/footer/footer.directive.js
```diff
@@ -8,9 +8,37 @@
...
+      vm.activate();
+    }
+
+    /** @ngInject */
+    function FooterController($q, Version) {
```
stackus: Remove `$q` now that is is not being used. The IDE should be pointing these out in the form of warning inspections (slim lines on the very right edge of the window)

sonejah21: Good tip! I'll be watching for these. I never noticed them before.

PR #916, opened 2015-09-25
added in the option of Manager to the user form

sonejah21:


src/client/app/components/user-form/user-form.html
```diff
@@ -114,6 +114,7 @@
                             required>
                         <option value="user">User</option>
                         <option value="admin">Admin</option>
+                        <option value="manager">Manager</option>
```
stackus: Place the manager option before the admin option. It had fewer permissions and we can use order to imply access levels along with the names.

PR #929, opened 2015-10-15
edited the Wizard 

sonejah21:
removes disabled buttons in footer for cleaner intro and finish using "ng-switch-when"; added intro message and congratulations message

### DIFF
--- a/src/client/app/components/wizard/wizard-modal.html
+++ b/src/client/app/components/wizard/wizard-modal.html
@@ -8,7 +8,7 @@
 <div class="wizard-modal__body" ng-switch="vm.state">
   <div ng-switch-when="intro">
     <div class="wizard-modal__question">Welcome to the Wizard!</div>
-    <div class="wizard-modal__subtitle">Determine the best-suited products for your project<br>by answering a set of simple questions.</div>
+    <div class="wizard-modal__subtitle">Determine the best-suited products for your project<br>by answering the following set of questions.</div>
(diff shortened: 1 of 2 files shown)


src/client/app/components/wizard/wizard-modal.html
```diff
@@ -7,7 +7,9 @@
 </div>
 <div class="wizard-modal__body" ng-switch="vm.state">
   <div ng-switch-when="intro">
-    <button type="button" class="btn-rounded wizard-modal__button" ng-click="vm.startWizard()">Start Wizard</button>
+    <div class="wizard-modal__question">Welcome to the Wizard!</div>
+    <div class="wizard-modal__subtitle">Determine the best-suited products for your project<br>by answering six simple questions.</div>
```
stackus: The number of questions is variable. We cannot display a count here in the text.

sonejah21: Yeah, I used the styling from the "wizard-modal__question" - didn't want it to necessarily act like a question, just match the same font size/style as the questions. I will duplicate the styling and give it a different name.

stackus: My comment was about this `by answering six simple questions`. Your reply about styling confuses me.

sonejah21: Oh that! Got it. I totally misunderstood you. Changing the correct part.

PR #980, opened 2015-10-21
Fixed save issues editing a  project and the back-to route

### LINKED ISSUE #942: Cannot save after editting a project
This issue may be fixed already.  Please confirm, or fix if needed.

voxlol:
Resolves #942 

There's a new issue where you can't save an edited project after loading the page because the form-validation looks like it wants the "Maintenance Day" field to be in a data format of YYYY-MM-DD instead of YYYY-MM-DD HH:MM:SS (which is default). 

You can bypass this issue atm by clicking on the calendar and selecting a date to remove the timestamp

See: https://github.com/projectjellyfish/api/issues/981


chriskacerguis:
@voxlol can you please make a new issue with the problem you found, then link it in this ticket?

PR #984, opened 2015-10-22
edited all buttons and page headers to conform to 'Add x' & 'Edit x'

sonejah21:
For consistency: Changed 'ADD A' and 'CREATE' to be just 'ADD'; same for 'EDIT A' to be 'EDIT'. This is in both buttons and page headers.


chriskacerguis:
Assuming that it passes, :+1:

stackus:
:+1:

PR #989, opened 2015-10-26
Fixed fisma sample question tag error

### LINKED ISSUE #940: Wizard results (filters) accumulate the more times you do the Wizard
Wizard results (filters) accumulate the more times you do the Wizard

@stackus - can we get more info on this bug?

voxlol:
Resolves #940


chriskacerguis:
:+1:

Issue #982, opened 2015-10-21
Form validation issues with "Maintenance Day" project question


voxlol:
#980 ticket describes the issue in detail

PR #990, opened 2015-10-26
Fixes maintenance day time bug

AllenBW:
screen shot 2015-10-26 at 4 05 50 pm
By modifying the projects seed, removing the time component from each maintenance day project question, editing projects will kickoff without a hitch.

Resolves: #982


stackus:
:+1: as long as the build passes

PR #993, opened 2015-10-27
Filter undefined project answers to fix ngRepeat issue

### LINKED ISSUE #985: ngRepeat fails for project answers on details page
- Create 2 new project questions.
- Click on the project details page of a project.
screen shot 2015-10-22 at 2 25 28 pm

Previous projects that don't yet have an answer will have an undefined answer id and break the ngRepeat

voxlol:
Closes #985


src/client/app/states/projects/details/details.state.js
```diff
@@ -95,13 +95,17 @@
...
+      vm.project.answers = lodash.filter(projectQuestions, hasAnswer);
 
       function addAnswer(question) {
         var answer = lodash.find(vm.project.answers, 'name', question.name);
 
         angular.extend(question, answer || {});
       }
+
+      function hasAnswer(answer){
+        return answer.value !== undefined;
```
stackus: `undefined` is undefined. Use `angular.isDefined()` or `angular.isUndefined()` instead of comparing to undefined.

Wouldn't this also throw out all questions that have no answer or default answer?

voxlol: Yea it should throw out all the questions that have no answer or default answer. Is that ok? I don't see the value of having a question listed with no answer, and it also protects against new questions added that automatically apply to all current standing projects.

stackus: Not really, if a question is not marked as required then the user might not have provided an answer. The question was still asked and is still a valid question so we'd want to display the question.

Wasn't the issue with that ngRepeat was failing for answers without an id? Should be tossing out objects without an `answer.id`. Could also change the track by to use `$index` instead of `answer.id`.

voxlol: That's a lot easier. I'll make a different pull request with the change to track by $index isntead

PR #1003, opened 2015-10-29
Add Order status and status_msg columns

### LINKED ISSUE #934: Orders have no status columns
Orders need to report their status. Orders start off in 'pending' and will advance to either 'complete', 'failed'.

Columns
status: integer
status_msg: string

Enum states for status
pending: initial state, set by the worker on failure if more retries exist
working: actively being worked by a worker
complete: final success state
failed: final failure state

voxlol:
Closes #934

### DIFF
--- a/db/migrate/20151029161157_add_status_to_orders.rb
+++ b/db/migrate/20151029161157_add_status_to_orders.rb
@@ -1,6 +1,6 @@
 class AddStatusToOrders < ActiveRecord::Migration
   def change
-    add_column :orders, :status, :integer
+    add_column :orders, :status, :integer, default: 0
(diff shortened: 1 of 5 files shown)


stackus:
It would be good to include the order status and message into the serialized output.

db/migrate/20151029161157_add_status_to_orders.rb
```diff
@@ -0,0 +1,6 @@
+class AddStatusToOrders < ActiveRecord::Migration
+  def change
+    add_column :orders, :status, :integer
```
stackus: Should provide a default of 0 for this column.

stackus:
:+1:

PR #1005, opened 2015-10-29
Adds order button and custom answer fields to service compare modal

### LINKED ISSUE #965: Compare modal is missing some parts
Modal is missing the following parts:

* Extra properties for each product being compared
* Restore order button

AllenBW:
Compare modal now supports displaying the contents of a products "answers" object and will properly align answers of similar names.  Additional, when the compare modal is accessed from the "add service" button on a project page, the order button will appear.  This presently redirect the user to the orders state.

closes #965 
screen shot 2015-10-30 at 12 21 37 am
screen shot 2015-10-30 at 12 21 42 am

### DIFF
--- a/src/client/app/components/compare/compare.service.js
+++ b/src/client/app/components/compare/compare.service.js
@@ -119,7 +119,7 @@
       function processProperties(product) {
         for (var idx = properties.length; --idx >= 0;) {
           data.properties[properties[idx]].values.push(
-            lodash.result(lodash.find(product.answers, {'name': properties[idx]}), 'value'));
+            lodash.result(lodash.find(product.answers, {name: properties[idx]}), 'value'));
         }
       }
 
--- a/src/client/app/components/compare/compare.service.js
+++ b/src/client/app/components/compare/compare.service.js
@@ -119,7 +119,7 @@
       function processProperties(product) {
         for (var idx = properties.length; --idx >= 0;) {
           data.properties[properties[idx]].values.push(
-            lodash.result(lodash.find(product.answers, {'name':properties[idx]}), 'value'));
+            lodash.result(lodash.find(product.answers, {'name': properties[idx]}), 'value'));
         }
       }
 
--- a/src/client/app/components/compare/compare.service.js
+++ b/src/client/app/components/compare/compare.service.js
@@ -81,16 +81,19 @@
   }
 
   /** @ngInject */
-  function CompareModalController(lodash, productList, project) {
+  function CompareModalController(lodash, productList, project, $modalInstance, $state) {
     var vm = this;
 
+    vm.project = project;
     vm.products = productList;
     vm.rowData = [];
+    vm.orderService = orderService;
 
     buildData();
 
     function buildData() {
       var properties = [];
+      var column = 0;
       var data = {
         description: [],
         setup: [],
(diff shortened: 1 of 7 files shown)


src/client/app/components/compare/compare.service.js
```diff
@@ -81,16 +81,19 @@
   }
 
   /** @ngInject */
-  function CompareModalController(lodash, productList, project) {
+  function CompareModalController(lodash, productList, project, $modalInstance, $state) {
     var vm = this;
 
+    vm.project = project;
```
stackus: Why is this being added to the vm? I can't find where it's being used in the template.

src/client/app/components/compare/compare.service.js
```diff
@@ -100,14 +103,17 @@
       };
 
       angular.forEach(productList, processBasics);
-      vm.rowData.push({name: 'Description', values: data.description});
       properties = lodash.uniq(properties.sort(), true);
       angular.forEach(properties, initProperty);
       angular.forEach(productList, processProperties);
```
stackus: All references to properties can be removed and replaced with the answers.

src/client/app/components/compare/compare.service.js
```diff
@@ -100,14 +103,17 @@
...
       angular.forEach(properties, appendProperty);
-      vm.rowData.push({name: 'Setup', values: data.setup});
-      vm.rowData.push({name: 'Hourly', values: data.hourly});
-      vm.rowData.push({name: 'Monthly', values: data.monthly});
+      lodash.forOwn(data, createRows);
```
stackus: Not sure I like the idea of dumping all of the properties of a product object out into the compare modal. We were specific about the rows on purpose.

src/client/app/components/compare/compare.service.js
```diff
@@ -115,6 +121,11 @@
         data.hourly.push(product.hourly_price);
         data.monthly.push(product.monthly_price);
         properties = properties.concat(lodash.keys(product.properties));
+
+        if (0 < product.answers.length) {
+          lodash.forEach(product.answers, addAnswerKey);
+        }
+        column++;
```
stackus: No need to track columns, see the properties handling code below.

src/client/app/components/compare/selected-products.directive.js
```diff
@@ -8,7 +8,9 @@
   function SelectedProductsDirective() {
     var directive = {
       restrict: 'AE',
-      scope: {},
+      scope: {
+        project: '='
```
stackus: Why is this being as a requirement of using a compare modal?

src/client/app/components/compare/compare.service.js
```diff
@@ -134,6 +145,20 @@
       function isPurchasable() {
         return angular.isDefined(project);
       }
+
+      function addAnswerKey(answer) {
```
stackus: Reuse the properties methods instead.

src/client/app/components/compare/compare.service.js
AllenBW: Order button or no order button

src/client/app/components/compare/compare.service.js
AllenBW: Wasn't sure what properties was so I left it alone, will get rid of it.

src/client/app/components/compare/compare.service.js
AllenBW: Again, didn't know who's it was so I left it alone, will get rid of it.

src/client/app/components/compare/compare.service.js
stackus: Project can be removed and instead of using a orderService function use `ui-sref="orders.create({productId: {whatever-gets-the-product-id})"` to send the user to the order page for the product.

Project should be removed as a variable from compare and selected-products.

src/client/app/components/compare/compare.service.js
AllenBW: Presently properties are not displayed, only the previously specified and
any answers that were provided

stackus:
The three prices should come last after the answers.

AllenBW:
Oh, an easy change to make, I had no idea.

AllenBW:
all pr feedback addressed, pictures updated to reflect changes.

src/client/app/components/compare/compare.service.js
```diff
@@ -123,17 +118,14 @@
 
       function processProperties(product) {
         for (var idx = properties.length; --idx >= 0;) {
-          data.properties[properties[idx]].values.push(product.properties[properties[idx]]);
+          data.properties[properties[idx]].values.push(
+            lodash.result(lodash.find(product.answers, {'name':properties[idx]}), 'value'));
```
stackus: I would think there would be a JSCS violation here since there isn't a space before `properties[idx]`

stackus:
:+1:

PR #1006, opened 2015-10-30
Add status and status_msg columns to orders list 

### LINKED ISSUE #1004: Display order status in UX
Now that status has been added to orders, #1003, we should display this information in the pages and in places where orders are being displayed.

voxlol:
Closes #1004 

I think it might be a good idea to colorize the 4 different status states to convey the information more effectively

### DIFF
--- a/src/client/app/states/orders/list/list.state.js
+++ b/src/client/app/states/orders/list/list.state.js
@@ -58,6 +58,7 @@
         order.product_name = order.product.name;
         order.total = order.total();
         order.staff = new Staff(staff).fullName();
+        order.status = order.status[0].toUpperCase() + order.status.slice(1);
 
         return order;
       }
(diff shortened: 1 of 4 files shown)


stackus:
Go ahead and leave the message our of the list, do not want to crowd the display.

There is a `status` directive which will add color for statuses.

stackus:
You'll also want to add the status, and this time status message, to the details screen for the order.

src/client/app/states/orders/list/list.state.js
```diff
@@ -58,6 +58,7 @@
         order.product_name = order.product.name;
         order.total = order.total();
         order.staff = new Staff(staff).fullName();
+        order.status = order.status[0].toUpperCase() + order.status.slice(1);
```
stackus: This looks like you're trying to `titlecase` the string. I believe lodash offers a function to do just that.

voxlol:
Where are the details screen for the order? When I click an order it brings me to a product page, which can be shared with multiple orders

stackus:
We'll track it down Monday.
:+1:

PR #1013, opened 2015-11-02
Adds UX for archiving projects

### LINKED ISSUE #935: UX for Archiving Projects
We need a UX for archiving projects.

FYI - Don't allow projects with Active services to be archived.

AllenBW:
screen shot 2015-11-03 at 8 04 41 am
screen shot 2015-11-03 at 8 02 46 am
screen shot 2015-11-03 at 8 03 06 am
screen shot 2015-11-03 at 8 02 39 am

Converts the "Edit" button to a group now including "Archive", said button is disabled when a project has services that are active.
Hijacks the delete project action to act as archive (rather than delete).
Updates relevant directives on the project.details state to suppress actions when a project is archived.

Items of additional note: 
* archived table doesn't show till there's cause
* confirmation modal gets wacky due to the source button being placed in the upper right, on the agenda, another pr to revert confirmation modal to be centered

closes #935

### DIFF
--- a/src/client/app/components/project-memberships/project-memberships.directive.js
+++ b/src/client/app/components/project-memberships/project-memberships.directive.js
@@ -10,6 +10,7 @@
       restrict: 'AE',
       scope: {
         projectId: '=',
+        projectArchived: '=?',
         memberships: '='
       },
       link: link,
--- a/src/client/app/components/project-memberships/project-memberships.directive.js
+++ b/src/client/app/components/project-memberships/project-memberships.directive.js
@@ -1,4 +1,4 @@
-(function() {
+(function () {
   'use strict';
 
   angular.module('app.components')
@@ -9,8 +9,7 @@
     var directive = {
       restrict: 'AE',
       scope: {
-        projectId: '=',
-        projectArchived: '=?',
+        project: '=',
         memberships: '='
       },
       link: link,
@@ -33,6 +32,7 @@
       vm.activate = activate;
       vm.showModal = showModal;
       vm.remove = remove;
+      vm.archivedProject = archivedProject;
 
       function activate() {
       }
@@ -82,6 +82,10 @@
         function deleteFailure() {
           Toasts.error('Could not remove membership. Try again later.');
         }
+      }
+
+      function archivedProject() {
+        return null === vm.project.archived ? true : false;
       }
     }
   }
--- a/src/client/app/components/project-memberships/project-memberships.directive.js
+++ b/src/client/app/components/project-memberships/project-memberships.directive.js
@@ -1,4 +1,4 @@
-(function () {
+(function() {
   'use strict';
 
   angular.module('app.components')
@@ -32,7 +32,7 @@
       vm.activate = activate;
       vm.showModal = showModal;
       vm.remove = remove;
-      vm.archivedProject = projectArchiveCheck;
+      vm.readOnly = readOnlyCheck;
 
       function activate() {
       }
@@ -84,7 +84,7 @@
         }
       }
 
-      function projectArchiveCheck() {
+      function readOnlyCheck() {
         return null === vm.project.archived ? true : false;
       }
     }
--- a/src/client/app/components/project-memberships/project-memberships.directive.js
+++ b/src/client/app/components/project-memberships/project-memberships.directive.js
@@ -32,7 +32,7 @@
       vm.activate = activate;
       vm.showModal = showModal;
       vm.remove = remove;
-      vm.archivedProject = archivedProject;
+      vm.archivedProject = projectArchiveCheck;
 
       function activate() {
       }
@@ -84,7 +84,7 @@
         }
       }
 
-      function archivedProject() {
+      function projectArchiveCheck() {
         return null === vm.project.archived ? true : false;
       }
     }
--- a/src/client/app/components/project-memberships/project-memberships.directive.js
+++ b/src/client/app/components/project-memberships/project-memberships.directive.js
@@ -85,7 +85,7 @@
       }
 
       function readOnlyCheck() {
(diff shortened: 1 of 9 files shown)


stackus:
Feedback based on screenshots

- "ARCHIVED" is mispelled as "ACHEIVED"
- The confirmation modal is half off the screen, directive accepts placement parameters.
- Archive and Edit have been "smushed" together.

src/client/app/components/project-memberships/project-memberships.directive.js
```diff
@@ -10,6 +10,7 @@
       restrict: 'AE',
       scope: {
         projectId: '=',
+        projectArchived: '=?',
```
stackus: Can this be done another way? Sending in this "status" is a code smell. (project memberships shouldn't require an archived boolean to function)

src/client/app/components/project-services/project-services.html
```diff
@@ -23,7 +23,10 @@
         <status type="{{ :: row.health }}">{{ ::row.health }}</status>
       </td>
       <td class="tables__cell tables__cell--actions">
-        <a ui-sref="services.details({serviceId: row.id})" class="tables__action" title="Edit">
+        <a ui-sref="services.details({serviceId: row.id})"
```
stackus: Code styling feedback

- Don't break up a tag into multiple lines unless it exceeds the maximum line length and then only break it up enough to fit. The IDE provides a code formatting shortcut that can be used to align to our code style.
- Put `ng-if` as the first attribute on a tag so a quick scan can determine if tags are toggling.

stackus:
Instead of passing in a new parameter called `projectArchived` which if left alone and later we needed to add an additional check to each directive we'd end up with logic checking multiple values. In fact we see this already happening with the services directive.

Another approach would be to pass in a function to test against if the directive should allow modifications or should allow xyz action to happen.

src/client/app/components/project-memberships/project-memberships.directive.js
```diff
@@ -32,6 +32,7 @@
       vm.activate = activate;
       vm.showModal = showModal;
       vm.remove = remove;
+      vm.archivedProject = archivedProject;
```
stackus: Refers to `archivedProject`.

src/client/app/components/project-memberships/project-memberships.directive.js
```diff
@@ -82,6 +83,10 @@
           Toasts.error('Could not remove membership. Try again later.');
         }
       }
+
+      function archivedProject() {
```
stackus: refers to `archivedProject`

src/client/app/components/project-services/project-services.directive.js
```diff
@@ -30,9 +30,14 @@
       var vm = this;
 
       vm.activate = activate;
+      vm.archivedProject = archivedProject;
```
stackus: ditto

src/client/app/states/projects/details/details.state.js
```diff
@@ -103,6 +105,45 @@
         angular.extend(question, answer || {});
       }
     }
+
+    function activeServices() {
+      var status = lodash.pluck(vm.services, 'status');
+      if (0 === status.length) {
+        return false;
+      } else {
```
stackus: Can remove the else branch.

src/client/app/states/projects/list/list.html
```diff
@@ -43,7 +43,47 @@
   </tr>
   </tbody>
 </table>
-
+<hr/>
+<region ng-if="::vm.projectsArchived.length > 1" heading="Archived Projects">
```
stackus: Use `>= 1` or ` > 0`.

src/client/app/states/projects/list/list.html
```diff
@@ -43,7 +43,47 @@
...
+    <table st-table="vm.projects" class="tables tables--bordered">
```
stackus: Tables in a region shouldn't have borders.

src/client/app/states/projects/list/list.html
```diff
@@ -43,7 +43,47 @@
...
+        </td>
+        <td class="tables__cell tables__cell--centered">
+          <status type="{{ ::row.health }}">{{ ::row.health }}</status>
+        </td>
+        <td class="tables__cell tables__cell--budget">{{ ::row.budget | currency }}</td>
+        <td class="tables__cell tables__cell--spent">{{ ::row.spent | currency }}</td>
+      </tr>
+      </tbody>
+    </table>
+  <div ng-if="::0 === vm.projects.length" class="tables__empty">
```
stackus: The region is never displayed if there are no archived projects, so this section will never be displayed.

src/client/app/states/projects/details/details.state.js
AllenBW: required to disable archive button for a project that has services that "might" be in use, unless we want to kill for now and wait for "active" var to come into play..

src/client/app/states/projects/list/list.state.js
```diff
@@ -36,6 +36,7 @@
     vm.projects = projects;
     vm.activate = activate;
     vm.title = 'Projects';
+    vm.projectsArchived = lodash.uniq(lodash.pluck(vm.projects, 'archived'));
```
stackus: Why are you having to use `uniq`?

Also I'd suggest using `lodash.partition` to generate the lists of projects. Should also allow you to remove several `ng-if` statements in the template as well.

AllenBW: we pluck the the list of "archived" status from the projects, if none are, unique will result in a single item array (because the array is full of nulls)

stackus:
:+1:

PR #1015, opened 2015-11-02
Fix hourly and monthly order pricing

### LINKED ISSUE #1014: Fix small bug with product pricing
Hourly row is displaying monthly price, and monthly price is displaying hourly price. Need to swap

voxlol:
Closes #1014


stackus:
:+1:

PR #1016, opened 2015-11-03
Enhancement/1001 data excludes

### LINKED ISSUE #1001: Use `except: [...]` to exclude attributes for actions that do not require a complete attribute set
*Update title after finding a workable solution*

Would like more control over what attributes are sent back for each endpoint. The first use case would be limiting the number of attributes sent back for lists. Lists, in most cases, could use a subset of a models attributes.

Either add an use `*CollectionSerializer` serializers, or discover a way to customize the attributes another way using existing serializers, or something else.

AnkyrinRepeat:
Closes #1001

Added attribute exclusion for Projects, Orders, Services, and Marketplace.

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -13,7 +13,7 @@ class OrdersController < ApplicationController
 
   def index
     authorize Order
-    respond_with_params orders
+    respond_with_params orders, index_respond_options
   end
 
   api :GET, '/orders/:id', 'Returns an order by :id'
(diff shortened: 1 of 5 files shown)


app/controllers/orders_controller.rb
```diff
@@ -46,6 +46,11 @@ def create
 
   private
 
+  def index_respond_options
+    { each_serializer: OrderSerializer, except: [:order_questions, :active, :updated_at, :staff_id,
```
stackus: Was `active` added to orders recently?

stackus:
:+1:

PR #1017, opened 2015-11-03
If user is already authenticated, redirects them to dashboard state

### LINKED ISSUE #939: Login Screen needs manual URL editing in order to login
The login screen remains stuck with username and password entered into fields until manually removing the `/login` from the URL.

After removing the '/login', it takes you to the dashboard and works just fine. 

See the attached video demonstrating this issue.

AllenBW:
If on the login state and already logged in, posting credentials returns a 302.  Documentation for $http states "A response status code between 200 and 299 is considered a success status and will result in the success callback being called."  This commit introduces an "isAuthenticated" call prior to submitting credentials. 

No image necessary as no views modified.

closes #939

@chriskacerguis this should solve the issue, lemme know if not.


src/client/app/components/login-form/login-form.directive.js
```diff
@@ -56,9 +56,13 @@
         showValidationMessages = true;
 
         if (vm.form.$valid) {
-          AuthenticationService.login(vm.email, vm.password)
-            .success(loginSuccess)
-            .error(loginError);
+          if (AuthenticationService.isAuthenticated()) {
+            loginSuccess();
+          } else {
```
stackus: Instead of branching with else return early.

stackus:
I'm not convinced this will resolve the issue described in #939. It also does not resolve the issue of the server sending 302s when a user appears logged in.

chriskacerguis:
Seems like the issue would be with the cookies?  Perhaps the cookie domain?

stackus:
In the case of the dev and test servers that's exactly what it sounds like is happening.

AllenBW:
Roger that, digging in.

AllenBW:
Ok, would like to propose that cookies are fine, the original issue was not reproducible following the below steps:

1. clear all jellyfish cookies 
2. start up `rails s -e development` log in on localhost:3000/login - results in success, then logout
3. kill `rails s -e development`
4. start up `rails s -e test` log in on localhost:3000/login - results in success

That being said, if one logsin on either test or development), skips logging out, then attempts to login again, while the _jellyfish_core_session attribute lastAccessed is updated, a 302 is thrown back (which to angular $http != success) and we are not redirected to the dashboard state.  

Could use a bit of guidance on  how to resolve the 302 issue experienced when a logged in user attempts to log in again (some sorta conditional in block starting at 16 of sessions_controller.rb?)

stackus:
A 302 is being sent back because of the `before_action` for the action in the `SessionsController`.

PR #1018, opened 2015-11-03
Fix counter_cache deprecation issue

### LINKED ISSUE #943: Fix sample data rake task - deprecation issue.
We should update the sample data rake task to resolve this deprecation issue.  I got several of these messages when running rake sample:demo

````
DEPRECATION WARNING: Automatic updating of counter caches on through associations has been deprecated, and will be removed in Rails 5. Instead, please set the appropriate `counter_cache` options on the `has_many` and `belongs_to` for your associations to groups_staff. (called from block (5 levels) in <top (required)> at /app/lib/tasks/sample.rake:100)
````

AnkyrinRepeat:
Closes #943
The counter_cache needs to be declared as part of the belongs_to association
dependent: :destroy fixes an issue where API DELETE destroy did not decrement staff_count.

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -13,6 +13,6 @@
 class Group < ActiveRecord::Base
   has_many :memberships
   has_many :projects, through: :memberships
-  has_many :groups_staff, counter_cache: :staff_count
-  has_many :staff, through: :groups_staff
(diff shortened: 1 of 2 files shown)


stackus:
:+1:

PR #1031, opened 2015-11-04
Permit Staff phone data to properly save to the DB

### LINKED ISSUE #1030: Permit Staff phone # data to save to database
Right now the phone # data in Staff creation doesn't properly save to the database

voxlol:
Closes #1030


stackus:
:+1:

PR #1036, opened 2015-11-05
Resolve 302 return on staff relogin 

### LINKED ISSUE #939: Login Screen needs manual URL editing in order to login

AllenBW:
If on the login state and already logged in, posting credentials returns a 302. Documentation for $http states "A response status code between 200 and 299 is considered a success status and will result in the success callback being called."

No image necessary as no views modified.

closes #939

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,7 +1,7 @@
 class SessionsController < Devise::SessionsController
   respond_to :json, :html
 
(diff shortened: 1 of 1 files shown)


stackus:
:+1:

PR #1051, opened 2015-11-09
Enhancement/1035-email-validation

### LINKED ISSUE #1035: Add unique email checking prior to submit on User Forms
See #969 bullet point 1

If I create a new user with an email in the database,
screen shot 2015-11-05 at 10 14 09 am
is the response sent back. And the UX has an error popup that isn't very useful

To avoid this, the user form should be able to asynchronously compare that the input email is unique and not currently in the database without submitting.

voxlol:
Closes #1035 

Wanted to add async validation for the email, instead it's validating on form submit.

### DIFF
--- a/app/models/staff.rb
+++ b/app/models/staff.rb
@@ -36,6 +36,9 @@ class Staff < ActiveRecord::Base
   acts_as_paranoid
   acts_as_taggable
 
+  scope :by_email, -> email { where :email => email }
+  scope :include_deleted, -> { :with_deleted }
+
   has_many :alerts, as: :alertable
   has_many :authentications
   has_many :memberships, through: :groups
--- a/app/models/staff.rb
+++ b/app/models/staff.rb
@@ -36,7 +36,7 @@ class Staff < ActiveRecord::Base
   acts_as_paranoid
   acts_as_taggable
 
-  scope :by_email, -> email { where :email => email }
+  scope :by_email, -> email { where email: email }
   scope :include_deleted, -> { :with_deleted }
 
   has_many :alerts, as: :alertable
--- a/app/models/staff.rb
+++ b/app/models/staff.rb
@@ -37,7 +37,6 @@ class Staff < ActiveRecord::Base
   acts_as_taggable
 
   scope :by_email, -> email { where email: email }
-  scope :include_deleted, -> { :with_deleted }
 
   has_many :alerts, as: :alertable
   has_many :authentications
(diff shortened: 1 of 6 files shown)


voxlol:
@stackus Code Climate found 1 new issue but it won't tell me what it is. When I click on the details it says "You don't have permission to access that".

stackus:
It doesn't like that you've added a new branch to the look up.

Look into using the `has_scope` gem so that you can reuse one of the existing branches.

has_scope: https://github.com/plataformatec/has_scope

app/models/staff.rb
```diff
@@ -36,6 +36,9 @@ class Staff < ActiveRecord::Base
   acts_as_paranoid
   acts_as_taggable
 
+  scope :by_email, -> email { where email: email }
+  scope :include_deleted, -> { :with_deleted }
```
stackus: This line if it were to remain should be `scope :include_deleted, -> { with_deleted }` but since `with_deleted` is already a scope* it is redundant and cane be removed.

\* : `with_deleted` is provided by the `acts_as_paranoid` gem.

src/client/app/components/forms/forms/user.form.js
```diff
@@ -124,6 +125,27 @@
...
+    }
+
+    function uniqueEmail(view, model, scope) {
+      var def = scope.q.defer();
+      scope.Staff.query({by_email: view}).$promise.then(handleRequest);
+
+      return def.promise;
+
+      function handleRequest(res) {
+        if (res.length > 0) {
```
stackus: Using the ternary operator here would eliminate the bracket baggage.

`res.length > 0 ? deferred.reject() : deferred.resolve();`

stackus:
Does this unique check cause any issues when editing a staff record?

voxlol:
Doh yea looks like thats the issue

voxlol:
Or not. I just made a fix and it's still failing without telling me why

chriskacerguis:
@voxlol see me tomorrow and I can get you access to the error.

stackus:
@voxlol Yes, I was referring to the `staffs` method when I said look up above.

Actually it's in the Staff model that CC is complaining about a new 'too much complex code outside of methods' violation. All I see added are two scope methods. One of those new scopes can be removed but I'm wondering if the remaining one will still be too much or CC to be happy.

As for that lookup method, it can be reduced back to the branch pair it had before and the `has_scope` updates can be added to the else branch.

And for your information CC complains about complex code in controllers a lot due to Apipie. I have an issue open looking for discussion for an alternative or suggestions on refactoring the code.

voxlol:
I got the email validation method to skip requests to the server if it passes the checks. I'll fix the lookup and push up the code. Thanks for all the feedback & help!

stackus:
:+1:

PR #1052, opened 2015-11-09
Adds service_output model to services & details state

### LINKED ISSUE #1038: Cannot access or display service outputs
Service may have many outputs created by the provisioning process or by other actions taken on a service.

- [x] Endpoint or accessible method for `service_outputs`
- [x] Display outputs on the service details page

AllenBW:
Provisioning output made available via 'includes' with param 'service_output'
Updated service.details to display provisioning output when available.
Updated sample.rake and services.yml to handle service_output model.

closes #1038 
screen shot 2015-11-09 at 3 21 58 pm
screen shot 2015-11-09 at 3 21 38 pm

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -23,7 +23,7 @@ class Service < ActiveRecord::Base
 
   has_many :alerts, as: :alertable
   has_many :latest_alerts, -> { latest }, class_name: 'Alert', as: :alertable
-  has_many :service_output
+  has_many :service_outputs
 
   has_one :order
   has_one :project, through: :order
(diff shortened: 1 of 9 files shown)


stackus:
The `services.yml` file has some odd formatting and the newline at the end is missing. Also, could you include some other `value_types` to demo and test with.

app/models/service.rb
```diff
@@ -23,7 +23,7 @@ class Service < ActiveRecord::Base
 
   has_many :alerts, as: :alertable
   has_many :latest_alerts, -> { latest }, class_name: 'Alert', as: :alertable
-  has_many :service_outputs
+  has_many :service_output
```
stackus: Why was this made singular?

stackus:
:+1:

PR #1055, opened 2015-11-10
Add tests for provider model

### LINKED ISSUE #949: RSPEC Failed: Provider add some examples to (or delete) /home/travis/build/projectjellyfish/api/spec/models/provider_spec.rb
````
  3) Provider add some examples to (or delete) /home/travis/build/projectjellyfish/api/spec/models/provider_spec.rb
     # Not yet implemented
     # ./spec/models/provider_spec.rb:25
````

AnkyrinRepeat:
Closes #949.

Added tests to confirm model is storing data and assigning providers a type.

### DIFF
--- a/spec/models/provider_spec.rb
+++ b/spec/models/provider_spec.rb
@@ -22,5 +22,13 @@
 require 'rails_helper'
 
 RSpec.describe Provider, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+  it 'assigns a type' do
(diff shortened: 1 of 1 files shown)


stackus:
These tests refer to `id`s that have been set elsewhere in the system. If and when other tests are added these `id`s will no longer point to the data you expect.

Use FactoryGirl to create data to test with. Every test should be able to run on it's own.

stackus:
:+1:

PR #1057, opened 2015-11-11
Ux/1034 phone validation

### LINKED ISSUE #1034: Need phone field validation on user forms
Right now there are no validations for phone numbers, and they are stored as a string up to 30 characters in the DB. I could enter "hello" on the form and it'll be saved in the DB because there's no validation at the moment

AnkyrinRepeat:
Closes #1034 

Adds client and server validation for staff phone numbers.


stackus:
:+1:

PR #1065, opened 2015-11-16
Adds core service actions functionality including sample deprovision

### LINKED ISSUE #1039: Services cannot be 'deprovisioned'
Services should have a counterpart action to `provision` to deprovision a service. Deprovisioning a service should be handled by the same class that provisioned the service.

- [ ] Add deprovision method
- [ ] Add action to UX

AllenBW:
screen shot 2015-11-16 at 9 23 38 am
Adds service_action_controller to manage the results of a PUT on `services/:id/actions/:operation`. At this time no action is taken on the service (to be handled by custom service classes).
Adds action button to service.details.  Presently, button button actions not pulled from api.

closes #1039


app/use_cases/create_service_order.rb
```diff
@@ -73,7 +73,7 @@ def save
   end
 
   def provision_service
-    service.delay.provision
+    service.operations('provision')
```
stackus: Don't make this change at this time. It may get refactored later.

If this change is made every single product/service will break.

app/models/service.rb
```diff
@@ -54,11 +54,7 @@ def self.policy_class
     ServicePolicy
   end
 
-  def actions
-    []
-  end
-
-  def provision
+  def operations(operation)
```
stackus: Want to point out the purpose of the two methods you're removing here.

- `actions` is meant to return a list of actions that can be taken. It will return an array of them.
- `provision` is the one action that all services will have, it'll be kicked off when an order is complete.

When you want to pass along an action for the the service to take you will need to use a different method other than the two above. Suggestions are `do_action`, or `handle_action`, or `start_action`.

app/models/service.rb
```diff
@@ -54,7 +54,10 @@ def self.policy_class
     ServicePolicy
   end
 
-  def actions
+  def operations(operation)
```
stackus: A few things here.

- `actions` should remain, call `operations`, and have a deprecation notice. See the ProductType class for an example of adding a deprecation note
- `operations` should return the empty array
- The new 'do the thing method' cannot be either one of those, it will need to be a new method name. The above are meant to return a list of available actions not do any work.

app/policies/service_policy.rb
```diff
@@ -11,6 +11,10 @@ def all_count?
     true
   end
 
+  def operations
```
stackus: You're missing the `?` here. The policy methods have the name pattern of `{action}?`.

In the policy files methods without a `?` all do something internal. Most with a `?` will be for actions. The exceptions are the private methods for checking a users access level.

app/models/service.rb
```diff
@@ -54,10 +54,18 @@ def self.policy_class
     ServicePolicy
   end
 
-  def actions
+  def operations
     []
   end
 
+  def start_operations(_opeation)
```
stackus: I think using the singular `start_operation` for the method makes more sense.

The parameter to the method is also mangled.

stackus:
:+1:

PR #1068, opened 2015-11-17
Add logging to Services and display on Service details state

### LINKED ISSUE #1037: Services need records of logs and errors
During the provisioning process it's possible some important messaging or errors might be returned by the provider or client and we should record these messages.

These logs should be readonly, and they don't need any methods or interfaces for a user to create them.

- [x] Store messages, record log level and timestamp
- [x] Display messages on the service details page
- [ ] Paginate results for UX and sort by newest first

voxlol:
Closes #1037


app/controllers/logs_controller.rb
```diff
@@ -0,0 +1,20 @@
+class LogsController < ApplicationController
+  def index
+    respond_with_params logs
+  end
+
+  def create
```
stackus: Logs cannot be created by users, so there isn't any need for this endpoint.

app/controllers/logs_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  private
+
+  def log_params
+    params.permit(:log_level, :message, :loggable_type, :loggable_id)
+  end
+
+  def logs
+    @_logs ||= query_with Log.where(loggable_type: params[:loggable_type], loggable_id: params[:loggable_id]), :includes, :pagination
```
stackus: `:includes` won't be needed I should think. Queries for logs will be simple tasks and getting the related models would be thrashy on the db.

AllenBW:
can you attach a ss of the update to service.details?

app/models/log.rb
```diff
@@ -0,0 +1,18 @@
...
+#  loggable_type :string
+#  loggable_id   :integer
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#
+
+class Log < ActiveRecord::Base
+  belongs_to :loggable, polymorphic: true
+
+  enum log_level: { ok: 0, warning: 1, critical: 2 }
```
stackus: Add on `debug` and `info`. Inject them in before `ok` and increment the integers. Nothing is using these values yet so there isn't any reason we can't keep them in some sort of logical order right now.

db/migrate/20151113182253_create_logs.rb
```diff
@@ -0,0 +1,14 @@
...
+  def change
+    drop_table :logs
+
+    create_table :logs do |t|
+      t.integer :log_level
+      t.string :message
+      t.string :loggable_type
+      t.integer :loggable_id
+
+      t.timestamps null: false
```
stackus: By placing the `t.timestamps null: false` at the top of the create table block we get those columns placed first in the column order. Later if additional columns are added the 'data' columns are all grouped together.

It makes no difference to the database or the application but when you're looking at the raw data you can better ignore the columns that don't matter.

src/client/app/resources/log.factory.js
```diff
@@ -0,0 +1,28 @@
...
+  angular.module('app.resources')
+    .factory('Log', LogFactory);
+
+  /** @ngInject */
+  function LogFactory($resource) {
+    var Log = $resource('/api/v1/logs/:type/:id' , {id: '@id', type: '@type'}, {
+
+    });
+
+    Log.defaults = {
```
stackus: Logs won't be created by the users so all of the defaulting and `newLog` can be removed.

To also make it clear to other developers that they should not attempt to create/update records add the following to this function

```
delete Log.get;
delete Log.save;
delete Log.remove;
delete Log.delete;
```

Removing those also keeps the random passerby from even attempting the make the call.

src/client/app/states/services/details/_details.sass
```diff
@@ -0,0 +1,10 @@
++block(logs-list)
+
+  +element(time)
+    width: 200px
+
+  +element(level)
+    width: 100px
+
+  +element(message)
```
stackus: You're going to want to allow one column to be without a width. This will allow the row to expand to fill the screen but by having only one column this way you can still control alignment.

The `message` column makes the most sense to be without a width since the messages could be really long.

db/schema.rb
```diff
@@ -168,6 +168,15 @@
   add_index "groups_staff", ["group_id"], name: "index_groups_staff_on_group_id", using: :btree
   add_index "groups_staff", ["staff_id"], name: "index_groups_staff_on_staff_id", using: :btree
 
+  create_table "logs", force: :cascade do |t|
+    t.integer  "log_level"
+    t.string   "message"
```
stackus: `message` could be really long so the type `text` should be used.

Quick difference of some column types.
`string` limits the value to 255 characters
`text` limits the value to typically 2gb

Both can accept a `limit: n` value to customize the number of characters further.

src/client/app/states/services/details/details.html
```diff
@@ -20,3 +20,31 @@
...
+      <td class="tables__cell logs-list__time">{{ ::log.created_at | date : 'medium' }}</td>
+      <td class="tables__cell logs-list__level">{{ ::log.log_level }}</td>
+      <td class="tables__cell logs-list__message">{{ ::log.message }}</td>
+    </tr>
+    </tbody>
+
+    <tfoot>
+      <tr>
+        <td colspan="5" class="text-center">
+          <div st-pagination="" st-items-by-page="15" st-displayed-pages="50"></div>
```
stackus: Does this paginate on the server or in the client?

voxlol:
@AllenBW attached a screenshot to show the added logging table
screen shot 2015-11-17 at 9 56 11 am

src/client/app/states/services/details/details.html
```diff
@@ -20,3 +20,31 @@
     <div class="details-table__detail">{{ ::output.value }}</div>
   </div>
 </details-table>
+
+<region heading="Logs">
```
stackus: Because logs are polymorphic that makes a great reason for the creation of this as a component.

voxlol:
Pushed up current progress. Working on server-side pagination with smart-table

stackus:
:+1:

PR #1069, opened 2015-11-17
Bug/1048 memberships not added

### LINKED ISSUE #1048: Memberships cannot be added to projects
The membership appears to be added successfully but when you refresh the page it disappears.

AnkyrinRepeat:
Closes #1048.

Creates/destroys group_staff association when creating a new group and updating a group. Orders groups ascending by id so order isn't changed by updating them.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -47,7 +47,7 @@ class GroupsController < ApplicationController
   def update
     authorize group
     group.update! group_params
-    #update_group_staff
+    update_group_staff
     respond_with_params group
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/groups_controller.rb
```diff
@@ -61,12 +65,29 @@ def destroy
...
+    end
+  end
+
   def group_params
     params.permit :name, :description, :staff_ids
   end
 
   def groups
-    @_groups ||= query_with Group.all, :includes, :pagination
+    @_groups ||= query_with Group.order('id').all, :includes, :pagination
```
stackus: Why are you adding an order by `id` here?

stackus:
Does this change keep the `staff_count` up to date as staff are added and removed?

app/controllers/groups_controller.rb
AnkyrinRepeat: Otherwise it orders the groups by updated_at, so if you change a group it gets dropped to the bottom of the list.

AnkyrinRepeat:
screen shot 2015-11-17 at 4 46 52 pm

Staff Count is updated when you add/subtract members from a group.

stackus:
:+1:

PR #1070, opened 2015-11-18
Feature/move notifications into core

mafernando:
Brings notifications into core, replicating behavior from the `jellyfish-notification` gem.

Default behavior is to have SMTP enabled set to false, but that Setting can be edited under Admin > Settings. 

Have verified this implementation works with Postmark as the SMTP server with my API Token specified as the user name and password.


config/initializers/listeners.rb
```diff
@@ -0,0 +1,11 @@
+require 'wisper'
+
+Rails.application.configure do
+  config.action_mailer.delivery_method = :smtp
```
stackus: Mail setup typically lives in the environment files, and I think both of these settings are already being set. Check the environment files and confirm and set there please.

Would also want to keep the initializers to a single task.

app/listeners/smtp_listener.rb
```diff
@@ -0,0 +1,10 @@
+class SMTPListener
+  def project_create(project, recipient)
+    SMTPMailer.project_create(project, recipient).deliver_later if smtp_enabled
```
stackus: Couple things here that could be improved upon.

- Using a mailer with the name of `ProjectMailer` would give a better 'read' of the scope of mails being sent. All mail would fall under a `SMTP` scope, that's the protocol after all, but not every mail would fall under a `Project` scope.
- Using mail method names like  `new_project_notification`, `thing_reminder` helps color in the mind what the user is getting. Duplicating the event name isn't very colorful.
- The two methods appear to be just about the same, it'd cut down on duplicate mailer templates to use a single method. If you determined the recipients list before hand and passed that in as the second argument it should work with minor changes.
- Return early or before getting to the sending of emails instead of checking each time you're about to send an email. By returning early it becomes clear that no more work is needed if an application feature is not enabled. Now in the future if other actions, other than mailings, needs to happen then wrapping the pair of calls in an if/unless would work.

app/controllers/projects_controller.rb
```diff
@@ -50,7 +50,7 @@ def show
   def create
     authorize Project
     project = Project.create project_params
-    publish(:publish_project_create, project, current_user) if project.persisted?
+    publish(:project_create, project, current_user.email) if project.persisted?
```
stackus: The entire `current_user` object should be passed in here. Passing in only the email kinda makes the event limited in purpose to sending emails imo.

app/models/setting/smtp.rb
```diff
@@ -0,0 +1,19 @@
+class Setting
+  class SMTP < Setting
+    def self.load_defaults
+      return unless super
+      transaction do
+        [
+          set('smtp_enabled', description: 'SMTP Enabled', value_type: :boolean, default: false),
+          set('smtp_address', description: 'SMTP Server IP Address', value_type: :string, default: 'address'),
+          set('smtp_port', description: 'SMTP Server Port', value_type: :string, default: 'port'),
+          set('smtp_user_name', description: 'SMTP User Name', value_type: :password, default: 'user_name'),
```
stackus: Did you mean for this field to be `value_type: :password`?

app/models/setting/smtp.rb
```diff
@@ -0,0 +1,19 @@
...
+          set('smtp_password', description: 'SMTP Password', value_type: :password, default: 'password'),
+          set('smtp_authentication', description: 'SMTP Authentication Method', value_type: :string, default: 'plain'),
+          set('smtp_default_sender', description: 'SMTP Default Sender', value_type: :string, default: 'sender@domain.com'),
+          set('smtp_default_recipient', description: 'SMTP Default Recipient', value_type: :string, default: 'recipient@domain.com')
```
stackus: What is the purpose of a default recipient? It's not immediately clear why one poor soul gets all the emails destined for others. The `email` for users is a required field so it shouldn't ever be nil. Can you explain the idea behind sending emails to this poor guy instead of just not sending the mail at all?

config/initializers/listeners.rb
mafernando: delivery method was already being set elsewhere, removed these.

addressed by 7f3f6a2

app/listeners/smtp_listener.rb
mafernando: These suggestions seem reasonable.

- Renamed SMTPMailer by f00e022
- Made method names more explicit by f00e022
- Consolidated templates by 510b4b4
- Returns early instead of checking against each mail by f00e022

app/controllers/projects_controller.rb
mafernando: Passing along current_user from projects controller by 510b4b4

app/models/setting/smtp.rb
mafernando: Per your feedback, changed this to :string by 24fa3cd

app/models/setting/smtp.rb
mafernando: Could not think of a good use case for default recipient, removed the setting by 510b4b4

stackus:
:+1:

PR #1073, opened 2015-11-20
Replace after_save with after_commit

### LINKED ISSUE #947: Change Rails after_save callbacks into after_commit
The top reason is we will be using workers and job queues a lot for processing things such as emails, provisioners, and so on.

This article will help with explaining what, why, and covers some gotchas that may be encountered.
http://www.justinweiss.com/articles/a-couple-callback-gotchas-and-a-rails-5-fix/

AnkyrinRepeat:
Closes #947


stackus:
:+1:

PR #1074, opened 2015-11-23
Improve Project budgets

### LINKED ISSUE #1064: Improve project budget reporting
Belongs to #1061 

Projects allow the setting of a project budget only. They should allow for a monthly budget as well so that budget overruns can be better reported.

Also the processes required to keep the projects budget usage needs to be updated too. Are there other options other than using Cron to keeping this data fresh? Can it be done in a lazy fashion?

voxlol:
Closes #1064 

- Adds `monthly_budget` column to the Project table
- Adds `monthly_budget` field to the Project form
- Orders can only be successfully placed if it doesn't exceed the monthly budget of the Project

### DIFF
--- a/app/use_cases/create_service_order.rb
+++ b/app/use_cases/create_service_order.rb
@@ -17,6 +17,9 @@ class CreateServiceOrder
   class PersistError < UseCase::Error
   end
 
+  class BudgetError < UseCase::Error
+  end
+
   attr_reader :order
 
   def initialize(current_user, order_params)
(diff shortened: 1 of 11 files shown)


app/use_cases/create_service_order.rb
```diff
@@ -45,6 +48,10 @@ def validate
     unless params[:service]['name'].present?
       fail UnnamedService, 'A name for the service was not given.'
     end
+
+    unless (product.monthly_price + (product.hourly_price * 750)) + project.monthly_spend <= project.monthly_budget
```
stackus: Could you make a `monthly_cost` method on `Product` to put the first half of this line into. It'll help clear up what is happening on this line. I think `Order` has a `monthly_cost` method that can be copied directly.

voxlol: Should be updated now, I copied the method over from the Order model

voxlol:
Add pass/fail test for monthly_budget overruns

stackus:
:+1:

Issue #1062, opened 2015-11-12
Cart: Allow multiple products to be ordered at once


stackus:
Belongs to #1061 

Orders should allow for several products to be ordered at one time. All of the products will belong to a single project.

Ordering should have the following validation and error handling

- Exceeds budget, either the project total budget, or the projects monthly budget
- Missing answer data for required questions that do not have defaults.

AnkyrinRepeat:
New format for POST requests should be:
```
{
  "project_id": 63,
  "products": [
    {
      "product_id": 23,
      "service": {
        "name": "Configured"
      },
      "answers": [
        {
          "id": 13,
          "name": "operating_system",
          "value": "Amazon Linux",
          "value_type": "string"
        },
        ...
      ]
    }
    ...
  ]
}
```

AllenBW:
@AnkyrinRepeat  is there branch i can test this out on?   Also, orders in the previous implementation dealt with the param `staff_id`... still a requirement?

AnkyrinRepeat:
I pushed it to the branch orders/1062-multiple-products-order. It currently will accept multiple orders and create records with them but I haven't fixed the response so it will return a 500 Error. 

I haven't actually touched any part of the 'create_service_order' code so assuming staff_id was required before it should be required still.

I don't think we can use whitelisting with an object of objects, is there a reason an array would be undesirable?

AllenBW:
Sweet, will hunt it down. 

Only reason array would be undesirable is because how the cart is built out, uses order id's as cart keys, just means we do a transformation on the frontend before sending it back.  Not a deal breaker, can be done.

PR #1075, opened 2015-11-27
Adds multi-project oriented, multi-state, product cart with enhanced budget reporting

### LINKED ISSUE #1061: Reimplement the cart
The cart should make a return so that users can do the following.

- [x] As a user I can add products under a project to the cart
- [x] As a user I can remove products from my cart
- [x] As a user I can see a list of products per project to see the impact to my budget should I order them all
- [x] As a user I can order all products in my cart for a project
- [x] As a user I can see the change in monthly costs should I add all products to a project
- [x] As a user I can see budget warnings should I add products which exceeds the project budget
- [x] As a user I must configure products before ordering

One catch is that products can require additional input from the user prior to ordering. So we must require that the products are configured prior to ordering.

- [ ] #1064 : Improve project budget reporting

**API**
- [x] #1062 : Allow multiple products to be ordered at once

**UX**
- [x] #1063 : Bring back the cart modal

\#TODO link sub-issues here

AllenBW:
Adds the following checkout reviews: budget, configuration, empty.
Adds cart configuration options list, presently includes reset default project.
Adds cart project helper for selecting default project on which to add a product.
Adds cart configuration helper for configuring products prior to  checkout.
Depending on state origin, able to add products to either origin, specified, or set default project.
Able to add products via marketplace, compare, and project details.
Able to remove individual products or whole projects.

Requires #1062 completion to fulfill #1061

closes #1061 

Recent commit updates modal size, adds column to indicate if product is configure, adds option to clear all projects from cart. (see below).  No other changes made.
screen shot 2015-12-02 at 4 47 57 pm

screen shot 2015-11-27 at 10 10 26 am
screen shot 2015-11-27 at 10 10 46 am
screen shot 2015-11-27 at 10 12 59 am
screen shot 2015-11-27 at 10 13 40 am


src/client/app/components/cart/_cart-modal.sass
```diff
@@ -0,0 +1,83 @@
++block(cart-modal)
```
stackus: A bunch of these styles will not be needed now. There's been some refactoring of the CSS for modals since the carts removal.

AllenBW: absolutely!  sweet. been using the globals in most places, not sure if we wanted to keep/use the legacy or not...

src/client/app/components/cart-project-helper/cart-project-helper.directive.js
```diff
@@ -0,0 +1,95 @@
...
+            options: projects,
+            valueProp: 'id',
+            labelProp: 'name'
+          }
+        },
+        {
+          key: 'projectDefault',
+          type: 'checkbox',
+          templateOptions: {
+            label: 'Default Project',
```
stackus: I think it would look nicer if the labels were swapped here. So `label` can be empty and the other one filled in.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    function clear() {
+      CartService.clear();
+    }
+
+    function checkout() {
+      var defaultProject = CartService.defaultProject();
+      var projectObject = CartService.items[defaultProject];
+      if (projectObject.orderCount === projectObject.itemCount) {
```
stackus: It is not immediately clear what this check is for. 

- *Why would an `orderCount` be different than an `itemCount`?*
- *Wouldn't each project only have one `orderCount`?*

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,235 @@
...
+        if (projectObject.project.budget >= projectObject.total) {
+          var projectOrder = {};
+          projectOrder.staff_id = SessionService.id;
+          projectOrder.project_id = defaultProject;
```
stackus: Seems odd that the default project is being assigned to each order. Wouldn't a `project_id` already exist for each projects grouping of products? And wouldn't you want to use that instead of forcing everything to be ordered under the default project?

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,235 @@
...
+      CartService.clear();
+    }
+
+    function checkout() {
+      var defaultProject = CartService.defaultProject();
+      var projectObject = CartService.items[defaultProject];
+      if (projectObject.orderCount === projectObject.itemCount) {
+        if (projectObject.project.budget >= projectObject.total) {
+          var projectOrder = {};
+          projectOrder.staff_id = SessionService.id;
```
stackus: The user making the purchase can be determined server side and this will not longer be needed.

src/client/app/components/cart/cart-service.factory.js
AllenBW: Order count is incremented when an product is added to a projects' `orders` object.  Item count is incremented when a product is added to the cart (the `products`) object.   To answer the first question, a product isn't added to `orders` list until it has been configured, so order count would differ from item if you have 10 items in the cart and configured only 2.

stackus: It would help make things more clear if the values being check were named better. It would also help to wrap up some of this logic into a method so it can be removed from the checkout method.

- Suggestion, `checkout` should be doing less.
- Suggestion, instead of managing a counter use a flag on each product item and provide a method to provide the boolean check if all items have been configured.
    - If you have to manage a separate counter you'll need to be sure it's kept up to date when products are removed and added. If other functions get added to the cart there's a chance this management would be forgotten and then a bug will creep in.

src/client/app/components/cart/cart-service.factory.js
AllenBW: Default project is invoked in and assigned in line 168 in order to determine which project a user is actively working on (i.e. which one they intend to submit for an order).  Because of changes made in #1062 we no longer keep project ids in a product object, so it can't be pulled from a grouping of products.  Because the checkout button is excluded from the project tabset, it requires context as to which project an order should be placed.  This is derived from the default project (which is updated, in addition to a number of other ways, whenever a user switches project tab, so we are making the assumption that a user desires to order a project they are actively looking at).  Another alternative would be to pull it from 'projectObject', which is created using the value from 168 so its kinda a wash. 

***tl;dr*** default project employed here to intuit the specific  project a user wants to order as checkout is project tab agnostic

stackus: Is the cart not organized as a collection of projects with related products?

If not then I'd suggest it be organized in that fashion. You would then be able to use a `currentTabIndex` property on the cart or wherever it is most appropriate to then be able to go back and get to that tabs project and products for any purpose. The index value would be changed/set when the user changes or clicks on a tab.

AllenBW: Cart is organized as such, a collection of projects (where the project id
is the key).   What you described is already preformed by `
CartService.items[defaultProject];`

src/client/app/states/marketplace/marketplace.state.js
```diff
@@ -50,12 +50,13 @@
...
   /** @ngInject */
   function StateController(logger, $q, VIEW_MODES, CatalogService, Tag,
-                           Compare, TAG_QUERY_LIMIT, $stateParams, WizardService) {
+                           Compare, TAG_QUERY_LIMIT, $stateParams, WizardService, $previousState) {
     var vm = this;
 
     vm.title = 'Marketplace';
     vm.tags = [];
     vm.viewMode = VIEW_MODES.list;
+    vm.projectId = null || $previousState.get().params.projectId;
```
stackus: If the application is loaded here at this url, what does `$previousState` contain? It'd be null I'd think and the application will encounter an error.

`null || anything` will never assign `null` at best the browsers compiler will optimize out the `null ||` bit.

stackus: The marketplace shouldn't be concerned with setting and holding onto a `project.id`.

src/client/app/states/marketplace/marketplace.html
```diff
@@ -21,4 +21,4 @@
   </div>
 </content-header>
 <catalog-category ng-repeat="category in vm.catalog" category="category" view-mode="vm.viewMode"
-                  required-tags="vm.tags"></catalog-category>
+                  required-tags="vm.tags" project-id="vm.projectId"></catalog-category>
```
stackus: Passing along the chosen project should not be the job of the marketplace or the catalog directive.

AllenBW: Directives for some reason don't like to seem to play along with pulling
state params, ( the reason for passing it in), because the function of the
add to cart changes when we have a projectId vs when we dont (throw pick a
project modal or add the product straight to the cart)

src/client/app/components/cart/cart-modal.html
```diff
@@ -0,0 +1,125 @@
...
+      <tab-heading>
+        {{ ::project.projectDetails.name}} ({{ project.productCount }})<a ng-click="vm.remove(project.projectDetails)" href=''> <i class="fa fa-remove"></i></a>
+      </tab-heading>
+
+
+      <alert ng-if="::0 >= project.projectDetails.budgetRemainder()" type="danger">
+        <strong>Budget Exceeded</strong> &mdash; The project budget has been exceeded!
+      </alert>
+
+      <alert ng-if="::project.projectDetails.monthly_spend && project.projectDetails.scheduleRemaining() > project.projectDetails.monthsRemaining()"
```
stackus: This logic and what's happening on this line would be clearer if it was happening in a function.

You might want to consider having a processing step in `activate()` that generates these alerts prior to display, then looping over them here in the template.

src/client/app/components/cart/cart-modal.html
```diff
@@ -0,0 +1,125 @@
...
+      </div>
+
+      <hr/>
+
+      <table st-table="orders" class="tables">
+        <thead>
+        <tr>
+          <th class="tables__heading cart-table--product">Product</th>
+          <th class="tables__heading cart-table--setup_price">Setup Price</th>
+          <th class="tables__heading cart-table--price">Monthly Price</th>
```
stackus: This should say `Monthly Cost`. We're displaying to the user the `hourly * 750 + monthly` prices.

src/client/app/components/cart/cart-modal.html
```diff
@@ -0,0 +1,125 @@
...
+        </thead>
+        <tbody>
+        <tr ng-repeat="line in project.products track by line.product.cartId">
+          <td class="tables__cell">
+            <product-description product="line.product"
+                                 link-to="products.details({productId: {{line.product.id}}})"></product-description>
+          </td>
+          <td class="tables__cell">{{ ::line.product.setup_price | currency}}</td>
+          <td class="tables__cell">
+            <computed-monthly-price pricing="line.product"></computed-monthly-price>
```
stackus: This directive might be deprecated now, I think product has a monthly_cost attribute now. Not 100% sure about this fact.

stackus: OK, I'm guessing you're not seeing a `monthly_cost` attribute on the product.

src/client/assets/sass/_modals.sass
```diff
@@ -40,6 +40,18 @@
     right: 30px
     transform: translateY(-50%)
 
+  +element(settings)
```
stackus: I did a search for `modals__settings` in this PR and it came back with nothing.

src/client/app/states/marketplace/marketplace.state.js
```diff
@@ -50,7 +50,7 @@
 
   /** @ngInject */
   function StateController(logger, $q, VIEW_MODES, CatalogService, Tag,
-                           Compare, TAG_QUERY_LIMIT, $stateParams, WizardService) {
+                           Compare, TAG_QUERY_LIMIT, $stateParams, WizardService, $previousState) {
```
stackus: Is this still needed now that `addService()` is being used on the `projects.details` screen?

src/client/app/components/confirmation/confirmation.directive.js
```diff
@@ -77,7 +77,7 @@
 
       var modalOptions = {
         templateUrl: 'app/components/confirmation/confirmation.html',
-        windowTemplateUrl: 'app/components/confirmation/confirmation-window.html',
+        windowTemplateUrl: 'app/components/common/modal-window.html',
```
stackus: This appears to be an unrelated change to the task. What's the purpose of the change?

src/client/app/components/cart/project-helper.service.js
```diff
@@ -0,0 +1,91 @@
...
+    activate();
+
+    function activate() {
+      initFields();
+    }
+
+    function addToCart() {
+      $modalInstance.close();
+      var selectedProject = lodash.find(projects, {'id': vm.selections.projectId});
+      if (vm.selections.projectDefault) {
```
stackus: This can be a single line.

`SelectedProjectHelper.selectProject(selectedProject, vm.selections.projectDefault);`

src/client/app/components/cart/project-helper.service.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  /** @ngInject */
+  function ProjectHelperController(projects, product, $modalInstance, CartService, SelectedProjectHelper, lodash) {
+    var vm = this;
+
+    vm.selections = {
+      projectId: null,
+      projectDefault: null
```
stackus: May as well default this to `false`.

src/client/app/components/cart/product-helper.service.js
```diff
@@ -0,0 +1,82 @@
...
+
+  /** @ngInject */
+  function ProductHelperService($modal) {
+    var service = {
+      showModal: showModal
+    };
+
+    return service;
+
+    function showModal(project, productRow) {
```
stackus: `project` doesn't appear to be used at all in this directive.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,200 @@
...
+      var projectOrder = {};
+      projectOrder.products = [];
+      var selectedProject = SelectedProjectHelper.selectedProject.id;
+      var projectObject = CartService.projects[selectedProject];
+      var allProductsConfigured = lodash.filter(projectObject.products, emptyName);
+
+      if (lodash.isEmpty(allProductsConfigured)) {
+        if (projectObject.projectDetails.budget >= projectObject.total) {
+          projectOrder.project_id = selectedProject;
+          lodash.forEach(projectObject.products, buildOrderList);
```
stackus: Using `lodash.map` here would mean you'd be able to assign the results directly to `projectOrder` instead of including it and modifying it in the function below.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,200 @@
...
+    function clear() {
+      CartService.clear();
+    }
+
+    function checkout() {
+      var projectOrder = {};
+      projectOrder.products = [];
+      var selectedProject = SelectedProjectHelper.selectedProject.id;
+      var projectObject = CartService.projects[selectedProject];
+      var allProductsConfigured = lodash.filter(projectObject.products, emptyName);
```
stackus: Use `lodash.all()` to return a boolean. `all()` takes a test function which will short circuit the checks if it finds a false result.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    function totalUpProject(project) {
+      if (angular.isUndefined(service.projects[project.id])) {
+        return;
+      }
+
+      service.projects[project.id].total = 0;
+      angular.forEach(service.projects[project.id].products, computeProductTotal);
+      function computeProductTotal(line) {
```
stackus: The JavaScript checkers should be saying you have an issue here. This needs a newline before the function.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,200 @@
...
+      };
+
+      var modal = $modal.open(modalOptions);
+    }
+
+    function inCart(project, product) {
+      return service.projects[project.id] && service.projects[project.id].products[product.cartId];
+    }
+
+    function totalUpProject(project) {
```
stackus: `totalUpProject()` should accept either an object representing the cart object you want totaled or the project id of the part you want totaled.

Using `lodash.reduce` instead of `angular.forEach` would be a better choice. With `reduce` you can avoid the assignment part of the work until it's all said and done.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,200 @@
...
+          name: ''
+        }
+      };
+      service.projects[projectId].productCount = lodash.size(service.projects[projectId].products);
+      totalUpProject(SelectedProjectHelper.selectedProject);
+      service.projects[projectId].products[cartId].product.cartId = cartId;
+      Toasts.toast(product.name + ' has been add to your cart under ' + SelectedProjectHelper.selectedProject.name);
+    }
+
+    function remove(project, productObj) {
```
stackus: I suggest using a different method for removing projects vs removing products.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,200 @@
...
+          productCount: 0
+        };
+      }
+
+      service.itemCountTotal += 1;
+
+      var cartId = lodash.size(service.projects[projectId].products) + 1;
+      service.projects[projectId].products[cartId] = {
+        product: lodash.cloneDeep(product),
+        price: 0,
```
stackus: Not seeing `price` being used. Is it?

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,200 @@
...
+      add: add,
+      remove: remove,
+      clear: clear,
+      isEmpty: isEmpty,
+      showModal: showModal
+    };
+
+    return service;
+
+    function add(project, product) {
```
stackus: `add()` feedback

- The method accepts a project, but then down on 25 it uses the information from `SelectedProjectHelper`. This is confusing and not clear to me why there's two methods of project info lookup happening. Wouldn't the two be the same?
-The cart model needs some simplification. You're holding onto and looking up a lot of ids.
  - If you're looping over things elsewhere you should use an array. Hashes do not guarantee order so tabs could jump in some browsers
  - You can use `lodash` to make a simple check if a project exists and also get to it in the cart with `lodash.find(model, {id: projectId})`
- `cartId` Instead of adding and tracking another id, use `ng-repeat`s `$index` from the view to reference the products in the products array
  - Use an array for the products collection, same hash reason as above

Two reasons to avoid adding and tracking additional ids to to lower the amount of mental juggling someone needs to do and it will reduce bugs creeping in as changes are made.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,200 @@
+(function() {
+  'use strict';
+
+  angular.module('app.components')
+    .factory('CartService', CartServiceFactory);
+
+  /** @ngInject */
+  function CartServiceFactory(Toasts, $modal, lodash, SelectedProjectHelper) {
+    var service = {
+      projects: {},
```
stackus: The cart model should be private, to make things a little clearer and to actually make it private you could move this out and make it a `var`.

src/client/app/components/cart/cart-modal.html
```diff
@@ -0,0 +1,125 @@
...
+
+  <p ng-if="vm.isEmpty()" class="cart-modal__empty">Your cart is empty.</p>
+</div>
+
+<div class="modals__footer">
+
+  <button type="button" class="btn-rounded modals__button--cancel" ng-click="$dismiss()">
+    Cancel
+  </button>
+  <button type="button" class="btn-rounded cart-modal__checkout"
```
stackus: Would this whole opening tag fit on a single line?

src/client/app/components/cart/cart-modal.html
```diff
@@ -0,0 +1,125 @@
...
+          <td class="tables__cell">
+            <computed-monthly-price pricing="line.product"></computed-monthly-price>
+          </td>
+          <td class="tables__cell tables__cell--actions">
+            <a class="tables__action" href="#" title="Remove" ng-click="vm.remove(project.projectDetails, line)">
+              <img src="/images/ic_deletequestion.png" alt="Remove"/>
+            </a>
+
+            <a class="tables__action" href="#" title="Configure"
+               ng-click="vm.configure(project.projectDetails, line)">
```
stackus: Would this whole tag fit on a single line?

src/client/app/components/cart/cart-service.factory.js
AllenBW: yep, was totally being used, but can be slimmed down/eliminated.

src/client/app/components/confirmation/confirmation.directive.js
AllenBW: deduplication/consolidation of modal files, slightly related as the common modal-window template was referenced for cart helpers.

src/client/app/components/cart/cart-service.factory.js
AllenBW: going to do away with passing params to same function to be titled `calcuateProjectCost()`.

no need to pass params, when we always know the selected project and action is only ever going to be on a project selected.  

will reduce HUZZAH!

src/client/app/components/cart/cart-service.factory.js
AllenBW: I could use additional clarification on this suggestion.  The forEach is intended to pluck only the required information  from each product associated with a project, then push it object to the `projectOrder.

src/client/app/components/cart/project-helper.service.js
AllenBW: Expected an assignment or function call and instead saw an expression.  (W030)

stackus: All you're doing is passing along the boolean value contained within `vm.selections.projectDefault` into `SelectedProjectHelper.selectProject`. This violation makes no sense.

stackus: Wait, did you leave the `if () {  } else { }` bits? I meant the whole conditional statement with else branch could be made into a single line.

src/client/app/components/confirmation/confirmation.directive.js
stackus: The confirmation modal window used custom styles. Have you confirmed the confirmation modals continue to work like before in all cases?

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,205 @@
...
+    function removeAllProjects() {
+      CartService.clear();
+    }
+
+    function clear() {
+      CartService.clear();
+    }
+
+    function checkout() {
+      var projectOrder = {};
```
stackus:
```
var projectOrder = {
  products: []
};

File: src/client/app/components/cart/cart-service.factory.js [hunk 28]
```diff
@@ -0,0 +1,205 @@
...
+      projectOrder.products = [];
+      var selectedProject = SelectedProjectHelper.selectedProject.id;
```
stackus: I'm a little concerned that the helper is being used not only as an authority of what project the user has chosen to put products onto but also what tab is actively being viewed in the cart. This is going to cause issues down the road.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,205 @@
...
+      var projectObject = vm.projects[selectedProject];
+
+      if (lodash.every(projectObject.products, configuredProductCheck)) {
+        if (projectObject.projectDetails.budget >= projectObject.total) {
+          projectOrder.project_id = selectedProject;
+          lodash.forEach(projectObject.products, buildOrderList);
```
stackus: What I meant by using map.
```
projectOrder.products = lodash.map(projectObject.products, buildOrderItem);
```

See changes for `buildOrderItem` (was `buildOrderList`)

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,205 @@
...
+      function saveSuccess() {
+        CartService.remove(projectObject.project);
+        Toasts.toast('Order accepted.');
+      }
+
+      function saveError() {
+        Toasts.error('Could not place order at this time.');
+      }
+
+      function buildOrderList(cartProduct) {
```
stackus:
```
function buildOrderItem(item) {
  return {
    product_id: item.product.product_type_id,
    answers: item.product.answers,
    service: item.service
  };
}
```

src/client/app/components/cart/cart-service.factory.js
AllenBW: To clarify, the active tab is not set by the selected project, however selecting a tab will reassign the selected project (also set the newly project as the default).    It would feel natural that a project selected as default would be the active tab when the cart modal is called, but this does not exist yet. Totally welcome refactoring suggestions!

stackus: Right, the value is being set as the tabs change. You're using the 'selectedProject' from the helper to look up which order to build and make. This should be using the tab index to lookup the data in the cart, not using the helper to find an id and then using the data with that id.

If the user chooses a default project using the helper, opens the cart, clicks a tab other than the one he chose as default, closes it then adds a product. That product will be added to the wrong project, or at least not the one the user expected it to since they chose a project and set it as default.

stackus:
How are you showing the user that they've configured a product or have not?

AllenBW:
Right now an alert/"do not pass" is thrown when a user attempts to place an
order with unconfigured products (order is not placed) same if the proposed
order ruins the exceeds the budget.

src/client/app/components/cart/cart-service.factory.js
```diff
@@ -0,0 +1,208 @@
...
+        return total + (parseFloat(line.product.monthly_price) + ((parseFloat(line.product.hourly_price)) * 750));
+      }
+    }
+  }
+
+  /** @ngInject */
+  function CartModalController(CartService, ProductHelper, $modalInstance,
+                               Toasts, Order, lodash, SelectedProjectHelper, $scope) {
+    var vm = this;
+    $scope.Object = Object;
```
stackus: What is this for?

stackus:
Looks good, we will need to wait for the backend to complete before merging.

Issue #1063, opened 2015-11-12
Cart: Add cart modal


stackus:
Belongs To #1061 

The cart modal makes a return. Most of it can be salvaged from the git history. Some major differences be added, namely the configuring of products. The cart navigation item should also be brought back, it also can be salvaged from the git history.

The cart will need to be updated to support the new, new to the old cart, features from #1061.

Also,
- Ordering needs to be removed from it's new home and put back into the cart.
- Answering the `service_questions` must happen in some way in the cart.

***Working Branch:*** https://github.com/projectjellyfish/api/compare/master...AllenBW:enhancement_1063

AllenBW:
A quick question with respect to answering service questions, going to calling this "configuration".  Should the configuration  occur prior to or after purchasing the contents of a cart?  First instinct on ordering workflow detailed below, this is working off the assumption we only have products as they relate to one project in the cart at a time.

1. From marketplace, a user adds services to "Compare", the button at the of the bottom of the compare table reads "Configure".
2. User selects "Configure" for a service, is prompted to select a project (if not previously been selected, or the user's origin is not project.details) and answer configuration questions. Upon completion user selects "Add To Cart".
3. Product is added to cart and the cart now has a Project set, all other products added will be against this project ("Clear/Reset Cart" could be an button to quickly remove all services). The cart, in addition to displaying the current products (each of which will show a remove action), will display the change in monthly costs and relevant budget warnings.
4.  In the cart selecting "Order Products" will display list previewing of each products' present configuration (also provide the option to reconfigure each if desired),and a "Confirm Order" button, which upon selection will add the products to the project and the cart will be cleared.


Second workflow instinct is below, products grouped by multiple projects in the cart (does away with aforementioned assumption):

1. From marketplace, a user adds services to compare, the button at the of the bottom of the compare table reads "Add to Cart".
2. "Add to Cart" reveals a modal that prompts the user to select a  project under which to add the product (this happens each time).
3. Product shows up in the cart under a "Project Tab" each new project will add an additional tab. In addition to the products list (which will show a remove action), will display the change in monthly costs and budget warnings  will be displayed in a project tab.  
4. Prom the "Project Tab" the button "Configure Project Products" upon selection, will take the user to a configuration page wherein the configuration required for each product is listed. After entering all required fields for each product requiring configuration, the user may select  "Order Products".  And so they will be added to the project and the "Project Tab" will be removed. And there will be much rejoicing.

sonejah21:
[...]

PR #1076, opened 2015-11-30
Enhancement 1063

AnkyrinRepeat:
Combine branches for issue #1062 and #1063.

### DIFF
--- a/src/client/app/components/cart-project-helper/cart-project-helper.directive.js
+++ b/src/client/app/components/cart-project-helper/cart-project-helper.directive.js
@@ -1,0 +1,95 @@
+(function() {
+  'use strict';
+
+  angular.module('app.components')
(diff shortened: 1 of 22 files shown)


src/client/app/components/cart-project-helper/cart-project-helper.directive.js
```diff
@@ -0,0 +1,95 @@
...
+
+  /** @ngInject */
+  function CartProjectHelperFactory($modal, Project) {
+    var service = {
+      showModal: showModal
+    };
+
+    return service;
+
+    function showModal(projectId, product) {
```
stackus: The `projectId` shouldn't need to be a required input. The service/helper that is tracking the value should be able to provide the id for this.

- If a chosen project has not been made then there's no id to default to
- If a chosen project has been made then it can be gotten from the service and shouldn't be passed in.

PR #1083, opened 2015-12-01
Validate UX to be 508 compliant

### LINKED ISSUE #1072: UX needs 508 compliance
We need to ensure that the current Jellyfish UX is 508 compliant.  This is a high priority.  This should help with that task:

http://www.508checker.com/what-is-508-compliance

@stackus we need to prioritize this ASAP.

voxlol:
Closes #1072 

- Installs ngAria to automatically add a lot of accessibility metadata
- Adds alt tags for all images
- Adds a lot of ARIA roles to important tags
- Add proper roles to html tags following the http://www.w3.org/TR/wai-aria/rdf_model.png model
- Each form input has an accessibility label

The site needed a coloring overhaul to comply with 508 compliance. Generally, you need a 4.5 contrast ratio between two different colors so I used http://leaverou.github.io/contrast-ratio/ and tried to get as close to 4.5 as possible. To contrast with our light background, the font colors got darker.

Here are some screenshots of the UX change

#### Dashboard
screen shot 2015-12-02 at 11 24 33 am

#### Tables
screen shot 2015-12-02 at 11 25 56 am

#### Project Details
screen shot 2015-12-02 at 11 25 17 am

#### Tags (greens)
screen shot 2015-12-02 at 1 14 21 pm

#### Forms & Red Errors
screen shot 2015-12-02 at 11 26 10 am


stackus:
:+1:

src/client/app/components/forms/formly.config.js
```diff
@@ -8,6 +8,7 @@
         prefix: 'Jellyfish Fields:'
       }
     }))
+    .filter('removeSpaces', removeSpacesFilter)
```
stackus: Instead of adding a filter and then having it run for every template use the `id` for the field that is generated by formly

http://docs.angular-formly.com/docs/custom-templates

`id` will be available in the label wrapper as well.

src/client/assets/sass/_app_colors.sass
```diff
@@ -12,31 +12,35 @@ $app-color-dark-gray: $color-cape-cod-approx
 $app-color-orange: $color-coral-approx
 $app-color-pink: $color-radical-red-approx
 $app-color-light-pink: $color-froly-approx
+$app-color-dark-pink: $color-lava
 
 // Greens
 $app-color-dark-green: $color-green-crayola-approx
 $app-color-green: $color-mountain-meadow-approx
+$app-color-dark-green-2: $color-viridian
+$app-color-teal : $color-tropical-rain-forest-approx
```
stackus: Extra space

src/client/assets/sass/_colors.sass
```diff
@@ -33,11 +33,14 @@ $color-red: #FF0000
 $color-radical-red-approx: #F13B54  // Pink
 $color-froly-approx: #F1757B        // Light Pink
 $color-coral-approx: #F88954        // Orange
+$color-lava: #CF1020                // Dark Pink
 
 // Greens
 $color-green: #00FF00
 $color-green-crayola-approx: #23A27E    // Dark Green
 $color-mountain-meadow-approx: #1DC58E  // Green
+$color-viridian: #14855f // viridian
```
stackus: Alignment and Capitalization please.

stackus:
When you added the 'dark' variant colors to `_app_colors`, did any uses of the non darkened version remain is use anywhere?

For example you added `dark-pink` and `dark-blue` then went around and replaced uses of `app-pink` with `app-dark-pink`. Does any `app-pink` remain?

voxlol:
Nope there aren't anymore uses of `app-pink` and `app-blue`

stackus:
The reason for asking is because the purpose of `_colors` and `_app_colors` is to reduce the amount of replacements you need to make when a color changes. In the application you should always be using `$app-...` and in `_app_color` you'd make a single change and avoid having to do a massive search and replace.

src/client/app/components/forms/formly.config.js
voxlol: Cool! That was very useful and quick fix

voxlol:
- Simplified color code back to normal
- Removed the removeSpaces filter in formly templates to use autogenerated id instead
- Changed the green color to use the same hue/saturation as the current green but with modified lighting that hits 4.6 contrast ratio

sonejah21:
Looks good!

stackus:
:+1:

PR #1085, opened 2015-12-02
Feature/update aws gem

mafernando:
Adds EC2 product type and service with provision and deprovision using custom templates embedded in the module. 

Monthly budget was recently added and needs to be set to a valid value
in the demo data, otherwise budget restrictions prevent the order from
going through. Ad-hoc solution to allow provisioning.

Referencing the [develop](https://github.com/projectjellyfish/jellyfish-aws/tree/develop) branch fro Github for now, will deploy through
RubyGems when master is stable.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -110,6 +110,6 @@ gem 'activesupport-json_encoder'
 gem 'has_scope'
 
 # Jellyfish Extensions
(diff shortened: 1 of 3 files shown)


stackus:
:+1:

PR #1087, opened 2015-12-02
Fix 508 compliance issue when there are no CMS records

### LINKED ISSUE #1086: ARIA role bug when there are no CMS records
screen shot 2015-12-02 at 3 27 46 pm

voxlol:
Closes #1086 

Saw a minor bug on the deployed version. This should be a small fix to get it back to compliant


stackus:
:+1:

PR #1107, opened 2015-12-04
Theme setting storage

### LINKED ISSUE #1091: Theming : Settings storage
The theme settings can be saved and loaded from the server as json data.

Belongs to #1067

voxlol:
closes #1091 

creates endpoint `/themes` and has a json column `colors` whose values are hex format

`theme` model has
1. Name (str)
2. Description (text)
3. Config (JSON) 
  - every value should be in hex color format i.e. `#FFFFFF`


AllenBW:
@voxlol this is great!! I have a few thoughts to bounce off of you after dabbling in the process of building out a theme for jellyfish as is.  Really this is just the suggestion to add two field types to what you have already posted. In addition to the `name`, `color` and `description` I would like to request we add `class` and `background-color`.  

An example of an entry with these additional fields could be:

```{
name: "active nav bar",
description: "Styling for the active page on the left navigation bar. Text and background color.",
class: ".sidebar__item-body--active",
color: "FFFFFF",
background-color: "14855F"
}```


Color and background-color would be optional fields, some elements have both some have one or the other.

File: app/controllers/themes_controller.rb [hunk 1]
```diff
@@ -0,0 +1,61 @@
+class ThemesController < ApplicationController
+  after_action :verify_authorized, except: [:show]
+
+  def self.error_codes
+    error code: 404, desc: MissingRecordDetection::Messages.not_found
+    error code: 422, desc: ParameterValidation::Messages.missing
+  end
+
+  def self.document_params
```
stackus: Apipie provides a way to define attribute blocks.

https://github.com/Apipie/apipie-rails#dry-with-param_group

app/controllers/themes_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  document_params
+
+  def create
+    theme_create_or_update
+  end
+
+  api :PUT, '/themes', 'Updates existing site theme.'
+  document_params
+
+  def update
```
stackus: This call will fail. There is no `update?` check defined in the policy.

You should add tests to cover the requests you're adding.

stackus:
@AllenBW Could you explain your plan hinted by this sentence `Color and background-color would be optional fields, some elements have both some have one or the other.`

How are you envisioning the creation of a theme, updating one, and mapping of styles back to things working?

stackus:
@AllenBW @sonejah21 ?

stackus:
@voxlol Will you be coming back to this?

stackus:
@voxlol Are you coming back to this PR?

PR #1108, opened 2015-12-05
Update the internal structure of the formly questions field

stackus:
The questions field was using a model for each generated internal field. This made
it very difficult to create dependencies between fields.

The generated fields now use a shared model that lives in the generated form. The
parent model is made available on the new shared model at `$parent`.

 This only modifies the internal structure of the question fields data. The structure
 of the model and the structure of the data sent to the server remain unchanged.

 - Made the `angular-aria` version number match the other `angular-*` packages

@mafernando


mafernando:
works great, merge away :+1:

PR #1109, opened 2015-12-07
Update rhel.md

lremurphy:
Installation instructions included an irrelevant "iconv-devel" package.  The pg gem version needs to be 0.18.3.  It was necessary to run a "bower install" in order for "gulp build" to function correctly.


stackus:
:+1: 

Thanks for the updates!

PR #1111, opened 2015-12-08
Orders/1062 multiple products order

### LINKED ISSUE #1062: Cart: Allow multiple products to be ordered at once

AnkyrinRepeat:
Closes #1062 

Should wait until #1063 is also closed before merging.

### DIFF
--- a/app/use_cases/create_service_order.rb
+++ b/app/use_cases/create_service_order.rb
@@ -45,19 +45,26 @@ class CreateServiceOrder
       fail UnapprovedProject, "Project '#{project.name}' has not been approved."
     end
 
+    validate_name
+
+    unless total_cost + project.monthly_spend <= project.monthly_budget
+      fail BudgetError, 'Adding this service will exceed the Project\'s monthly budget.'
+    end
+  end
+
+  def validate_name
     params[:products].each do |product|
       unless product[:service]['name'].present?
         fail UnnamedService, 'A name for the service was not given.'
       end
     end
+  end
 
+  def total_cost
     products.each do |product|
       @total_cost += product[:monthly_price]
     end
-
-    unless @total_cost + project.monthly_spend <= project.monthly_budget
-      fail BudgetError, 'Adding this service will exceed the Project\'s monthly budget.'
-    end
+    @total_cost
   end
 
   def build_service(service, product_id)
@@ -65,7 +72,7 @@ class CreateServiceOrder
     service.status = :pending
     service.status_msg = 'Provisioning service...'
     service.product_id = product_id
-    return service
+    service
   end
 
   def build_order
--- a/app/use_cases/create_service_order.rb
+++ b/app/use_cases/create_service_order.rb
@@ -25,11 +25,11 @@ class CreateServiceOrder
   def initialize(current_user, order_params)
     @user = current_user
     @params = order_params
+    @total_cost = 0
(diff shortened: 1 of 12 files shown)


app/use_cases/create_service_order.rb
```diff
@@ -25,11 +25,11 @@ class BudgetError < UseCase::Error
   def initialize(current_user, order_params)
     @user = current_user
     @params = order_params
+    @total_cost = 0
```
stackus: You can remove this and see the comment for `total_cost` below

app/use_cases/create_service_order.rb
```diff
@@ -45,42 +45,75 @@ def validate
...
-  def build_service
+  def validate_name
+    params[:products].each do |product|
+      unless product[:service]['name'].present?
+        fail UnnamedService, 'A name for the service was not given.'
+      end
+    end
+  end
+
+  def total_cost
```
stackus: Consider using memoization with the following
```
def total_cost
  @total_cost ||= products.map { |p| p[:monthly_price] }.inject(&:+)
end
```
This will map each product down to a single integer value, then use inject with a Proc shortcut to sum everything together.

app/use_cases/create_service_order.rb
```diff
@@ -45,42 +45,75 @@ def validate
...
     service.type = service.class.to_s
     service.status = :pending
     service.status_msg = 'Provisioning service...'
+    service.product_id = product_id
+    service
   end
 
   def build_order
-    order_params = params.merge(
+    @setup_price = 0
```
stackus: You can leave the `@` off of variables if your intend to only use them locally. Using `@` will make an instance variable.

app/use_cases/create_service_order.rb
```diff
@@ -89,18 +122,27 @@ def project
     raise MissingProject, 'The associated project cannot be located.'
   end
 
-  def product
-    @product ||= Product.find params[:product_id]
+  def products
+    product_ids = []
```
stackus: If you want to use memoization here you should be wrapping the "work" into a `begin` and `end` along with the value return.
```
def products
  @products ||= begin
    product_ids = []
    params[:products].each do |product|
      product_ids.push(product[:product_id])
    end
    Product.find(product_ids)
  end
rescue ActiveRecord::RecordNotFound
  raise MissingProduct, 'The associated product cannot be located.'
end
```
But the method can be shortened and the need for a `begin` and `end` can be replaced with
```
@products ||= Products.find params[:products].map { |p| p[:product_id] }
```

stackus:
I realized there's a bit of a logic error in the `total_cost` method. It has to do with how ActiveRecord is going to fetch the data given an array of ids. Basically the array with duplicates will have those duplicate values tossed out, so if an order had any duplicate `product_ids` you're going to be short on the products count if you compared the lengths of it and the original params length.

`total_cost` is going to need to be updated to handle the situation where duplicate products might exist in an order.

stackus:
Looks good, let's merge this with the UX work into a new branch and test there.

PR #1112, opened 2015-12-08
Update jellyfish-aws gem to point to master

mafernando:
Points jellyfish-aws gem to master and adds back S3 and RDS alongside EC2.


stackus:
:+1:

Issue #960, opened 2015-10-19
Wizard results should look different than regular search


chriskacerguis:
We should consider adding a visual indication to the search results that are refined in the Services list after you complete the Wizard.

Without some difference, a person will wonder if the Wizard changed anything in the list. If they do see the results change, they will notice that all the Wizard did was add in tags to filter on, which hardly seems magical and Wizard-like.

Suggested changes:

Less intensive:
Add a colored banner (matching Wizard colors) to the top of the page or each category that tells you these are the results based on the Wizard.
Add another form of simple visual instead of a banner.

More intensive:
The banner could also allow you to "See All Services" in the page, or in the category you're on, in case you don't want the refined results.
Add stars to the services that are highly recommended to you based on your Wizard results (if we plan on adding in this logic).

sonejah21:
<img width="982" alt="screen shot 2016-01-07 at 12 24 41 pm" src="https://cloud.githubusercontent.com/assets/5269893/12178910/b8d6ccb4-b539-11e5-8f9f-c0c9a12fb8b1.png">
Added customized modal close functionality to allow for a Toast, indicating results are from Wizard.

PR #1126, opened 2016-01-07
#960 - Added in a toast upon Wizard Modal closing

sonejah21:
screen shot 2016-01-07 at 12 24 41 pm

This toast shows up upon closing of the Wizard Modal.

### DIFF
--- a/src/client/app/components/wizard/wizard-modal.html
+++ b/src/client/app/components/wizard/wizard-modal.html
@@ -27,7 +27,7 @@
   <div ng-switch-when="complete">
     <div class="wizard-modal__question">Congratulations!</div>
     <div class="wizard-modal__subtitle">The Wizard has come up with the best service options for you.</div>
-    <button type="button" class="btn-rounded wizard-modal__button" ng-click="$close(vm.tags)">
+    <button type="button" class="btn-rounded wizard-modal__button" ng-click="$close(vm.tags, wizardAlert())">
       Show Matching Products
     </button>
   </div>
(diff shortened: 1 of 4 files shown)


src/client/app/components/wizard/wizard-modal.html
```diff
@@ -27,7 +27,7 @@
   <div ng-switch-when="complete">
     <div class="wizard-modal__question">Congratulations!</div>
     <div class="wizard-modal__subtitle">The Wizard has come up with the best service options for you.</div>
-    <button type="button" class="btn-rounded wizard-modal__button" ng-click="$close(vm.tags)">
+    <button type="button" class="btn-rounded wizard-modal__button" ng-click="vm.wizardSuccess(tags)">
```
stackus: What is `tags` referencing here?

sonejah21: So I just removed the `tags` inside of `wizardSuccess()`, leaving it blank, and that also works - I'll check this removal in. Thanks!

stackus: The called function `wizardSuccess` is still expecting a parameter. You should remove that parameter as well, since you're no longer passing anything, and it wasn't being used.

sonejah21: Sweet, I have now removed that as well.

sonejah21:
May I merge this PR in?

stackus:
:-1:

PR #1171, opened 2017-02-22
Handle running sync jobs to providers to populate provider_data

amalagaura:
Adding migration for provider_data and syncjob
Adding advisory_lock gem so we can do application level locking in postgres to ensure no more than 1 job runs at a time. The alternative transactions require locking a row for updates which on a provider would not be advisable. Otherwise we would need another table just for this purpose.


db/structure.sql
```diff
@@ -274,6 +274,26 @@ CREATE TABLE project_requests (
 
 
 --
+-- Name: provider_data; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE provider_data (
```
stackus: use `, id: :uuid` here instead of defining the id column below.

db/structure.sql
```diff
@@ -274,6 +274,26 @@ CREATE TABLE project_requests (
...
+    id uuid DEFAULT uuid_generate_v4() NOT NULL,
```
stackus: Move the timestamps to the front of the table.

config/initializers/scheduler.rb
```diff
@@ -12,6 +12,12 @@
     end
   end
 
+  scheduler.every '10m', overlap: false do
+    Provider.where(connected: true).select(:id).each do |provider|
```
stackus: nitpick wold prefer using `provider_id` here.

app/jobs/provider_data/sync_job.rb
```diff
@@ -0,0 +1,16 @@
+class ProviderData < ApplicationRecord
+  class SyncJob < ApplicationJob
+    INTERVAL = 1.min
```
stackus: Move to `ProviderData` and rename to `SYNC_INTERVAL`.

app/jobs/provider_data/sync_job.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    def perform(provider)
+      # timeout_seconds as 0 so that it will try to lock once, if the job is running the block will be skipped
```
stackus: Create a service `provider_data/sync.rb` and move this into there and have the job call it.

app/jobs/provider_data/sync_job.rb
```diff
@@ -0,0 +1,16 @@
...
+      ProviderData.with_advisory_lock("provider-data-sync-#{provider.id}", timeout_seconds: 0) do
```
stackus: Also want to keep the users and system from scheduling multiple syncs.

A solution might be to also pass along the providers last_synced_at and in the service exit early if the value does not match.

stackus: While the solution doesn't keep them from scheduling them, it will cause any extras to exit without doing any work.

test/fixtures/provider_data.yml
```diff
@@ -0,0 +1,23 @@
+# Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
+
+one:
+  data_type: template
+  provider: cloud_forms
+  name:
```
stackus: These will need names, the column won't allow nulls.

app/jobs/provider_data/sync_job.rb
amalagaura: Yeah my system allows users to create multiple jobs but is designed to complete those jobs quickly. So if you have 1 or multiple workers on a queue, it will process them really fast. If you have multiple queued up and one queue, it will still finish the first and then the rest will not run because it is less than 1 min.

PR #1172, opened 2017-02-22
Handle running sync jobs to providers to populate provider_data

amalagaura:
Adding migration for provider_data and a SyncJob
Adding advisory_lock gem so we can do application level locking in postgres to ensure no more than 1 job runs at a time. The alternative transactions require locking a row for updates which on a provider would not be advisable. Otherwise we would need another table just for this purpose.


stackus:
Will also need the controller, route, and serializer.

config/initializers/scheduler.rb
```diff
@@ -12,6 +12,12 @@
     end
   end
 
+  scheduler.every '10m', overlap: false do
+    Provider.where(connected: true).select(:id).each do |provider|
+      ProviderData::SyncJob.perform_later provider
```
stackus: Send in only the providers id. Avoid serializing an ActiveRecord object into Sidekiq jobs.

config/routes.rb
```diff
@@ -26,6 +26,8 @@
     resources :provider_types, only: [:index, :show]
     resources :providers do
       resource :connection, only: [:update], module: :providers
+      resource :sync, only: [:create], module: :providers
+      resources :provider_data, only: [:index]
```
stackus: `resources :provider_data` should be at the top level.

app/services/provider_data/sync.rb
```diff
@@ -0,0 +1,29 @@
...
+    sanitize do
+      required(:provider_id, ApplicationRecord::Types::UUID).filled(format?: ApplicationRecord::Types::UUID_REGEXP)
+      optional(:force, :bool).maybe(:bool?)
+    end
+
+    def perform
+      # timeout_seconds as 0 so that it will try to lock once, if the job is running the block will be skipped
+      ProviderData.with_advisory_lock("provider-data-sync-#{model.id}", timeout_seconds: 0) do
+        # If we have last synced more than INTERVAL we will work, otherwise we will succeed silently
+        if params[:force] || model.last_synced_at.nil? || model.last_synced_at < Time.current - ProviderData::SYNC_INTERVAL
```
stackus: Instead of time gating things here check that the job should be run against some metric. Previous suggestion was to pass in the last_sync_at value and to exit early if the value does not match the records value.

There's other reason to avoid time gating, but one is during development a developer would need to alter the interval time.

amalagaura: So we can check the last updated_at from each provider? A concern is if there are no provider_data for the provider we might keep repeating the service/query if the job is repeatedly queued. Do we still need to check against the minimum interval in that case?

amalagaura: Or get rid of only `Time.current` and have that be passed in to the service for testing? I am just not sure how to avoid repeatedly running the service if there is no provider_data to supply the last updated_at. Or maybe I am not clear about this.

stackus: There will be two ways a sync will be started. Either the scheduler will kick off the job, or a job will be kicked by a user choosing to start a manual sync.

In both cases we'll pass in two values into the job `provider_id` and `last_synced_at`. The job will then in turn pass those two values into the service.

The service will do two checks. First is the new one. It'll compare `provider.last_synced_at` against `params[:last_synced_at]` if they're different then return e.g. `return unless a == b`. The second check you've got already with the advisory lock.

These two checks will ensure that a list of sync requests run very quickly back to back will only allow the first to complete. If that list is run slower or just very long then the tailing jobs would also do no work since the last_synced_at value they were requested to run with is no longer the correct value.

When we return to the service from calling `provider.sync_data` or whatever the method will be called it'll update the last synced at timestamp before exiting the advisory lock/block.

amalagaura: OK now I understand the `params[:last_synced_at]` logic. It makes sense, I'll add it in tomorrow.

app/services/provider/sync/create.rb
```diff
@@ -0,0 +1,9 @@
+class Provider < ApplicationRecord
+  module Sync
+    class Create < ApplicationService
```
stackus: This service will be called from a controller and should include the following.

- Sanitization (allow provider_id)
- Model, it should load the Provider and use the last_synced_at value from the record and not trust a user to provide the correct value.
- Policy, only admins are permitted to fire off syncs.

app/services/provider_data/search.rb
```diff
@@ -0,0 +1,11 @@
+class ProviderData < ApplicationRecord
+  class Search < ApplicationService
```
stackus: The `collection_model` method will need to be overridden to add a call to `query_with`. We also want to apply a default of `available: true` to the query, while still allowing the user to filter for `available=false`.

app/services/provider_data/sync.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def perform
+      return unless model.last_synced_at.to_i == params[:last_synced_at].to_i
+      # timeout_seconds as 0 so that it will try to lock once, if the job is running the block will be skipped
+      ProviderData.with_advisory_lock("provider-data-sync-#{model.id}", timeout_seconds: 0) do
+        # Possibly have another transaction block in which we mark everything inactive, then process results
+        # and create or update active provider_data
+      end
+    end
+
```
stackus: add `private`

app/services/provider_data/search.rb
stackus: The `collection_model` already does the `query_with` call part, what has to happen is the second part I mention. Adding a default for available only results unless the user has filtered otherwise.

amalagaura: I see an example with a default in cart/search but I am not sure how to allow it to be overridden. Also not sure of the difference between `policy_scope(super)` and just `super`

stackus: `policy_scope` will use the `FooBarPolicy::Scope` class to apply constraints to the query. `ApplicationPolicy::Scope` just returns the query unchanged so not every policy will define constraints or use `policy_scope`.

Override by defining the method locally `def collection_model ... end`. Or did I misunderstand the point of confusion?

amalagaura: OK that was helpful I see the Scope classes now.

I meant that I saw an example of an overridden collection_model! in cart/search. It overrides using `policy_scope`, but I think we don't need that as policies seem to be context related.  

I can add a condition for available, but can a query override it?

stackus: Right, `policy_scope` isn't what we want here. You can use `finalize_params!` to set the filter for available to true if it's not already passed in.

PR #1174, opened 2017-07-31
Bug/cleanup rubocop and reek violations

AnkyrinRepeat:
### Description

Add additional configuration to reek and refactor code to follow rubocop and reek suggestions.

### Motivation and Context
Refactor Project Jellyfish so it has zero reek/rubocop violations.

### How Has This Been Tested?
Used pre existing tests to ensure no change in functionality from refactor. 

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Adds database migrations (Requires all environments to run `rake db:migrate` to work)
- [ ] Adds new Gemfile dependencies (Requires all environments to run `bundle install` to work)
- [ ] Adds new ENV variables (description of ENV and purpose is required in the above description)

### Checklist:

- [x] My code follows the code style of this project.
- [x] My code has been cleaned of commented, logging, and inline test code.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

The following tests fail on both develop and this branch:
test_user_should_not_be_able_to_change_their_own_role 
test_admin_should_not_be_able_to_change_email_to_another_users_email 
test_create_remote_auth_session_for_new_user 
test_create_remote_auth_session_for_existing_user
test_authenticate_from_remote_user_params 
test_create_project_questions 
test_new_user_registration_via_remote_headers

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ AllCops:
     - bin/**
     - lib/tasks/sample.rake
     - lib/jellyfish_fog_aws/spec/dummy/**/*
+    - lib/manageiq_client/lib/manageiq-client.rb
     - test/**/*
   TargetRubyVersion: 2.3
 
@@ -54,3 +55,15 @@ Style/MultilineMethodCallIndentation:
 Style/FrozenStringLiteralComment:
   Enabled: false
 
+
+Metrics/CyclomaticComplexity:
+  Max: 8
+
+Metrics/PerceivedComplexity:
+  Max: 8
+
+Metrics/AbcSize:
+  Enabled: false
+
+Metrics/ParameterLists:
+  Max: 6
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,6 @@ AllCops:
     - lib/tasks/sample.rake
     - lib/jellyfish_fog_aws/spec/dummy/**/*
     - lib/manageiq_client/lib/manageiq-client.rb
-    - lib/goby/**/*
     - test/**/*
   TargetRubyVersion: 2.3
 
@@ -56,12 +55,16 @@ Style/MultilineMethodCallIndentation:
 Style/FrozenStringLiteralComment:
   Enabled: false
 
+Style/PredicateName:
+  NameWhitelist:
+    - has_one
+    - has_many
 
 Metrics/CyclomaticComplexity:
   Max: 8
 
 Metrics/PerceivedComplexity:
-  Max: 8
+  Max: 9
 
 Metrics/AbcSize:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,7 @@ AllCops:
     - lib/tasks/sample.rake
     - lib/jellyfish_fog_aws/spec/dummy/**/*
     - lib/manageiq_client/lib/manageiq-client.rb
+    - lib/goby/**/*
     - test/**/*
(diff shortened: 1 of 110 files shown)


AllenBW:
🔫  👉  🐕 💥

AnkyrinRepeat:
Rubocop objected to the 'Place the . on the previous line, together with the method call receiver' correction, went with the rubocop interpretation.

AnkyrinRepeat:
Hey @stackus, would you mind taking a look at this PR?

Thanks

.rubocop.yml
```diff
@@ -6,6 +6,7 @@ AllCops:
     - bin/**
     - lib/tasks/sample.rake
     - lib/jellyfish_fog_aws/spec/dummy/**/*
+    - lib/manageiq_client/lib/manageiq-client.rb
```
stackus: This is jellyfish code and it should be tested.

app/models/concerns/flag_options.rb
```diff
@@ -11,56 +11,60 @@ module FlagOptions
   module ClassMethods
     def flag(key)
       skey = key.to_s
-      define_method(key) { self[:flags].include? skey }
+      define_method(key) { flags.include? skey }
```
stackus: Think would break the metaprogramming that this module does.

Only way to know is to run the application and test the feature.

app/models/service.rb
```diff
@@ -80,7 +80,7 @@ def check_status
 
   # TODO: unused?
   def connected?
-    provider.connected? or (self.status_message = 'Actions cannot be taken while the service provider is disconnected' and false)
+    provider.connected? || (self.status_message = 'Actions cannot be taken while the service provider is disconnected' && false)
```
stackus: The uses of `or` and `and` were on purpose here.

app/policies/application_policy.rb
```diff
@@ -1,7 +1,7 @@
 class ApplicationPolicy
   attr_reader :context, :record
-  alias_method :user, :context
-  alias_method :object, :record
+  alias user context
```
stackus: Not sure this change is correct.

1. `alias` and `alias_method` work differently
2. You're no longer passing in symbols?

app/policies/application_policy.rb
```diff
@@ -45,11 +45,11 @@ def resolve
 
   private
 
-  def is_manager?
+  def manager?
```
stackus: The use of `is_` is to avoid overlapping policy methods. Exceptions should be made for these methods.

app/policies/application_policy.rb
AnkyrinRepeat: Rubocop prefers alias and using variables instead of keys, if this change is inappropriate I'll disable the cop.

> Use alias instead of alias_method in a class body.
>  Use alias user context instead of alias :user :context.

>

AnkyrinRepeat:
Running ManageIQ using the manageiq/manageiq:fine-3 docker image and following their docker config instructions (http://manageiq.org/docs/get-started/docker) and Google Cloud I was able to connect to ManageIQ from Jellyfish and create an instance but in Jellyfish it hangs at provisioning the machine after the provisioning ends and I'm unable to terminate the instance from Jellyfish on both develop and bug branch. Are there any up to date documents on how to configure Jellyfish?

